PR #177, opened 2015-01-07
Relax configure-logger! preconditions to take more than a file path string

stephenbrady:
In addition to file path, take java.io.File, java.net.URL, java.io.InputStream, or org.xml.sax.InputSource as allowed by the underlying JoranConfigurator.

- Rename configure-logger-via-file! to configure-logger! and surrounding function args accordingly.

See Jira TK-131.

### DIFF
--- a/src/puppetlabs/trapperkeeper/logging.clj
+++ b/src/puppetlabs/trapperkeeper/logging.clj
@@ -72,25 +72,26 @@
             (.toInt level))
        (.setLevel root level)))))
 
-(defn configure-logger-via-file!
-  "Reconfigures the current logger based on the supplied configuration
-  file."
-  [logging-conf-file]
-  {:pre [(string? logging-conf-file)]}
+(defn configure-logger!
+  "Reconfigures the current logger based on the supplied configuration."
+  [logging-conf]
+  {:pre [(#{String java.io.File java.net.URL java.io.InputStream org.xml.sax.InputSource} logging-conf)]}
   (let [configurator (JoranConfigurator.)
         context      (LoggerFactory/getILoggerFactory)]
     (.setContext configurator (LoggerFactory/getILoggerFactory))
     (.reset context)
-    (.doConfigure configurator logging-conf-file)))
+    (.doConfigure configurator logging-conf)))
 
 (defn configure-logging!
-  "Takes a file path which can define how to configure the logging system.
+  "Takes a file path, url, file, or InputStream which can define how to
+  configure the logging system.
+  
   Also takes an optional `debug` flag which turns on debug logging."
-  ([logging-conf-file]
-   (configure-logging! logging-conf-file false))
-  ([logging-conf-file debug]
-   (when logging-conf-file
-     (configure-logger-via-file! logging-conf-file))
+  ([logging-conf]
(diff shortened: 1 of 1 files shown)


cprice404:
This broke some tests, and might warrant having another test or two added to validate the expected behavior.  Will follow up on the Jira ticket.

stephenbrady:
Indeed it did. An obvious oversight on my part! I pushed a fix.

nwolfe:
@stephenbrady Thanks, this will definitely improve things! In an effort to ensure we preserve these changes, would you mind adding more tests for the new types that the function supports?

The existing tests are here: https://github.com/puppetlabs/trapperkeeper/blob/master/test/puppetlabs/trapperkeeper/logging_test.clj#L21

src/puppetlabs/trapperkeeper/logging.clj
```diff
@@ -72,25 +72,26 @@
...
 
-(defn configure-logger-via-file!
-  "Reconfigures the current logger based on the supplied configuration
-  file."
-  [logging-conf-file]
-  {:pre [(string? logging-conf-file)]}
+(defn configure-logger!
+  "Reconfigures the current logger based on the supplied configuration."
+  [logging-conf]
+  {:pre [(#{String java.io.File java.net.URL java.io.InputStream org.xml.sax.InputSource} (type logging-conf))]}
```
rlinehan: Another way of doing this would be

```clojure
  {:pre [(some #(instance? % logging-conf) (list String java.io.File java.net.URL ...))]}
```
This has typically been our preferred method of checking multiple types in a pre- or post-condition, since rather than checking that the type of the argument is matches exactly what is listed, it will instead check that the argument is an instance of the classes listed.

For example:

```clojure
(#{java.io.InputStream} (type (clojure.java.io/input-stream "path/to/logging-conf")))
=> nil
```
since `clojure.java.io/input-stream` creates a `java.io.BufferedInputStream` by default, which isn't explicitly listed in the set. However,

```clojure
(instance? java.io.InputStream (clojure.java.io/input-stream "path/to/logging-conf"))
=> true
```

Writing the precondition using `type` means that a `logging-conf` of type `BufferedInputStream` would cause this to raise an error, even though `doConfigure` will accept a `BufferedInputStream` without error.

KevinCorcoran: I don't think pre-conditions should artificially restrict what values a function can handle.  In this case, it feels too onerous to bother with pre-conditions, I think it'd be best to just delete it altogether.

stephenbrady: @rlinehan good point
@KevinCorcoran yes, I think it makes sense to drop the precondition entirely as well. I only submitted a pull request to expand the list because the original code just had `string?`, which just shouldn't be.

rlinehan:
I added a comment about our preferred method of doing this type of pre-condition. However, in this case I agree with @KevinCorcoran, and I think it would be better to remove the pre-condition entirely.

I'm going to go ahead and merge this PR since it is an improvement over the current state of this. I'll open up a follow up PR removing the precondition.

@stephenbrady thanks for the contribution!

PR #179, opened 2015-02-11
(TK-142) Add support for logback evaluator filters

cprice404:
This commit adds support for logback's EvaluatorFilters.  These
require a dependency that we didn't have before (janino), and
allow really powerful scripting constructs to be specified
in the logback config file.  This will be extremely useful for
support escalations, e.g. in cases where we need to prevent
a problematic log message from bloating the log files on disk.

### DIFF
--- a/dev-resources/logging/logback-evaluator-filter.xml
+++ b/dev-resources/logging/logback-evaluator-filter.xml
@@ -1,0 +1,37 @@
+<!-- This is only used as a test file, not for actually configuring logging during tests -->
+<configuration>
+    <appender name="F1" class="ch.qos.logback.core.FileAppender">
+        <filter class="ch.qos.logback.core.filter.EvaluatorFilter">
+            <evaluator>
+                <matcher>
+                    <Name>matcher</Name>
+                    <regex>should get filtered</regex>
+                </matcher>
+                <expression>matcher.matches(formattedMessage)</expression>
+            </evaluator>
+            <OnMismatch>NEUTRAL</OnMismatch>
+            <OnMatch>DENY</OnMatch>
+        </filter>
+        <filter class="ch.qos.logback.core.filter.EvaluatorFilter">
+            <evaluator>
+                <matcher>
+                    <Name>omgMatcher</Name>
+                    <regex>OMGOMG</regex>
+                </matcher>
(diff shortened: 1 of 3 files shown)


fpringvaldsen:
:+1: Tests passed locally.

project.clj
```diff
@@ -8,12 +8,14 @@
   :pedantic? :abort
   :dependencies [[org.clojure/clojure "1.5.1"]
                  [org.clojure/tools.logging "0.2.6"]
+                 [clj-time "0.5.1"]
```
jeffmccune: Why is the clj-time dependency included?

cprice404: We already had a transitive dependency on it, this just pins to a specific version.  For some reason IDEA started complaining that we had multiple conflicting transitive deps on it.

dev-resources/logging/logback-evaluator-filter.xml
```diff
@@ -0,0 +1,37 @@
...
+        <append>false</append>
+        <encoder>
+            <pattern>%d %-5p [%c{2}] %m%n</pattern>
+        </encoder>
+    </appender>
+
+    <root level="info">
+        <appender-ref ref="F1" />
+    </root>
+</configuration>
```
jeffmccune: No newline at end of file.

jeffmccune:
+1 other than my two minor comments.  The clj-time dependency doesn't seem necessary, but if it's a good idea to have then don't let it block merging this in.

dev-resources/logging/logback-evaluator-filter.xml
cprice404: is that a thing?

project.clj
jeffmccune: Ah, got it.

jeffmccune:
I'll fixup the whitespace error and merge this in now.

dev-resources/logging/logback-evaluator-filter.xml
jeffmccune: Yeah, http://stackoverflow.com/questions/729692/why-should-files-end-with-a-newline

jeffmccune:
Merged into master as bed722d9a75ffca4ccb89dd33808e5f12084f772

PR #186, opened 2015-04-02
Add support for YAML config files

cybem:


src/puppetlabs/trapperkeeper/config.clj
```diff
@@ -67,7 +71,7 @@
     [path]
     (mapcat
       #(fs/glob (fs/file path %))
-      ["*.ini" "*.conf" "*.json" "*.properties" "*.edn"])))
+      ["*.ini" "*.conf" "*.json" "*.properties" "*.edn" ".yaml" ".yml"])))
```
rlinehan: I think this needs to be `"*.yaml" "*.yml"`, since `fs/glob` matches files using globbing.

I don't think there's currently anything testing this code path (parsing yaml files in a directory when that directory is passed as the config). It would be good to add a yaml file to `dev-resources/config/mixeddir` or one of the other directories that gets parsed to ensure that this works.

rlinehan:
@cybem Thanks for the contribution, and the super quick response! Can you add a test of that behavior? It could just be a matter of adding a yaml file to `dev-resources/config/mixeddir` and adding the values to the tests at https://github.com/puppetlabs/trapperkeeper/blob/master/test/puppetlabs/trapperkeeper/config_test.clj#L97-L112 and https://github.com/puppetlabs/trapperkeeper/blob/master/test/puppetlabs/trapperkeeper/config_test.clj#L123-L134. Otherwise, this looks good!

cybem:
Hello, @rlinehan! I fixed it. Thank you.

rlinehan:
:+1: This looks good to me.

fpringvaldsen:
:+1: This looks good to me as well!

fpringvaldsen:
@cybem Thanks for the contribution!

PR #188, opened 2015-04-09
(TK-197) Update prismatic library dependencies

dparis:
The prismatic/plumbing and prismatic/schema library dependencies are
lagging behind the latest stable version. This will increasingly
cause conflicts in trapperkeeper projects. This patch updates the
dependencies to the latest stable versions.

* Update project.clj dependencies

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -10,8 +10,8 @@
                  [org.clojure/tools.logging "0.2.6"]
                  [clj-time "0.5.1"]
                  [puppetlabs/kitchensink ~ks-version]
-                 [prismatic/plumbing "0.2.1"]
-                 [prismatic/schema "0.2.2"]
+                 [prismatic/plumbing "0.4.2"]
+                 [prismatic/schema "0.4.0"]
                  [org.clojure/tools.nrepl "0.2.3"]
                  [org.clojure/tools.macro "0.1.2"]
                  [ch.qos.logback/logback-classic "1.1.1"]
@@ -48,7 +48,7 @@
              :uberjar {:aot [puppetlabs.trapperkeeper.main]
                        :classifiers ^:replace []}}
 
-  :plugins [[lein-release "1.0.5"]]
+  :plugins [[lein-release "1.0.5" :exclusions [org.clojure/clojure]]]
 
   :main puppetlabs.trapperkeeper.main
   )


dparis:
Even if the CI build passes cleanly, you'll likely want to wait for:
* https://github.com/puppetlabs/trapperkeeper-webserver-jetty9/pull/116
* https://github.com/puppetlabs/clj-http-client/pull/39

dparis:
Just FYI, I've used `lein install` to deploy the updated versions to my local repo. My app is now able to require the updated prismatic deps and run `lein tk` using the jetty9 server/router without crashing.

cprice404:
@aperiodic @pcarlisle @senior I am thinking about merging these soon, maybe tomorrow or so.  Any concerns?

@aperiodic I know that you mentioned that in the more recent versions of schema you were hitting some issue related to recursive schemas.  I have a project that uses a couple of recursive schemas and I just tried updating it to schema 0.4.0, and everything seemed to work OK for me.  Are you still thinking that it would cause some problems for the classifier schema if we bumped up the version?

cprice404:
@aperiodic @pcarlisle @senior I will interpret your silence as enthusiastic support.  :)

@camlow325 do you have any problem with me doing releases of tk/tk-j9/c-h-c tomorrow?  Also any opinion on semver?  I'm leaning towards z since these changes should be invisible to any downstream projects that aren't using schema, and would (hopefully) show up as a version conflict for those who are?

kbarber:
@cprice404 we have some compat issues with 0.4.0, but nothing we can't solve I think (macro->core renaming mainly).

cprice404:
@kbarber great, thanks.

kbarber:
@cprice404 https://github.com/puppetlabs/puppetdb/pull/1351 FYI

camlow325:
@cprice404 No concerns from me about doing releases of those projects with the new schema version in tow.

aperiodic:
@cprice404 when we try to update the schema dependency in the classifier, we can't even run our tests due to [a schema protocol error in the puppetlabs http client](https://gist.github.com/aperiodic/ac0f369189d9051344bc).

aperiodic:
I thought it might be a stale code issue, but `rm -rf target` does not alleviate the problem.

dparis:
@aperiodic I'm pretty sure you're seeing the same issue I fixed here:  https://github.com/puppetlabs/clj-http-client/pull/39

v0.4.0 doesn't seem to like passing a protocol binding directly as a schema. I replaced it with a protocol schema which fixed the crash and seems to pass the test suites.

I created a puppet JIRA issue with subtasks to track how these fixes depend on one another: https://tickets.puppetlabs.com/browse/TK-195

aperiodic:
@dparis it does look like that issue.

@cprice404 until y'all release a new version of `puppetlabs/http-client` that has that issue fixed, then we can't even identify what problems, if any, the classifier might have with schema 0.4.0

cprice404:
@aperiodic once we do that, it's going to be hard to put the toothpaste back into the tube.  I've pulled down @dparis 's PRs locally and done some `lein install`s with them to verify that the work with all of the new code I have in flight.  You might want to do the same; I know it'll take a few minutes to get everything set up, but I don't see any other way of validating that these changes aren't going to irrevocably break your stuff.

As of right now, my plan is to merge all 3 of these PRs and do the corresponding Z releases of the various projects tomorrow morning (Belfast time), unless I hear some objection from you or someone else between now and then.  I've gotten sign-off from the PuppetDB and Puppet Server teams.

cprice404:
Going to merge this and then add a few touch-up commits to sync with the latest versions of the other projects.

PR #191, opened 2015-06-02
(maint) Flush cli messages; send errors to stderr

rbrw:
Flush cli-error and cli-help messages in main before System/exit, since
the JVM doesn't necessarily flush on exit, and redirect cli-error messages
to stderr, to follow the normal command-line convention.

### DIFF
--- a/src/puppetlabs/trapperkeeper/core.clj
+++ b/src/puppetlabs/trapperkeeper/core.clj
@@ -155,16 +155,17 @@
   [& args]
   {:pre [((some-fn sequential? nil?) args)
          (every? string? args)]}
-  (try+
-    (-> (or args '())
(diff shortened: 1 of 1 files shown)


cprice404:
+1

KevinCorcoran:
:+1: Looks good to me.

@rbrw Out of curiosity, did you actually observe any cases of messages not getting flushed before shutdown?

rbrw:
@KevinCorcoran yes, and you can probably demonstrate it by altering a "lein new app foo" core.clj to include a (System/exit 1) right after the hello world.

rbrw:
@KevinCorcoran at least that demonstrates it here.

jeffmccune:
:+1:

PR #192, opened 2015-06-16
(maint) Preserve stack, etc. when rethrowing in main

rbrw:
Explicitly rethrowing the pending exception adds another wrapper,
whereas (throw+) just passes on the existing exception.

The relevant bit from the throw+ docstring:

  Within a try+ catch clause, a throw+ call with no arguments rethrows
  the caught object within its original (possibly nested) wrappers.

### DIFF
--- a/src/puppetlabs/trapperkeeper/core.clj
+++ b/src/puppetlabs/trapperkeeper/core.clj
@@ -166,6 +166,6 @@
        (case (without-ns (:type m))
          :cli-error (quit 1 (:message m) *err*)
          :cli-help (quit 0 (:message m) *out*)
-         (throw+ m)))
+         (throw+)))
      (finally
        (shutdown-agents)))))


cprice404:
Sorry, I'm confused on this... doesn't this obscure some of the actual stack trace from the real flow of control through the code?  Why is that better than preserving it?

rbrw:
Not sure, but what prompted this was that we'd added our own try/catch around main and were throwing a {:type ::foo :message ...} exception that we wanted to catch and handle.

If I recall correctly, the current rethrow approach means that you can't use the normal slingshot syntax to extract the :message because it's been wrapped in another layer by trapperkeeper.

cprice404:
OK... well, I guess I don't have too strong of an opinion on this, but wrapping `main` seems like kind of an edge case, and I think that we've historically tried to prevent allowing slingshot exceptions to bubble all the way up to the end user... because, as unpleasant as it is for a user to have to see a Java stacktrace, the slingshot exceptions seem even more cryptic when they make it all the way up to the shell.

That's my recollection, anyway.  Also, just in general, the idea of catching and re-throwing an exception without passing along the new stack info seems risky.  I don't think it'd be a huge deal here, since 99% of the time this will be the outermost call in the stack anyway, but as a general thing it makes me kind of uneasy.

I could go either way on this change.  Maybe it'd be good to see if @senior , @KevinCorcoran , @camlow325 , or @pcarlisle have any thoughts.

rbrw:
I suppose at the moment I'd lean toward the transparent rethrow since TK's explicitly deciding that the exception isn't something it's interested in.

But more broadly, that's led me to think a bit about possible enhancements to TK, and I've filed a PDB ticket along those lines (to think about it a bit more): https://tickets.puppetlabs.com/browse/PDB-1628  

It's not mentioned there, but I've also pondered adding support for something like a :configuration-error, perhaps containing file :file and :line information, that could produce friendly top-level messages when a service detects a problem with the configuration (no idea if that's reasonable yet).

camlow325:
I don't have a super-strong opinion on this one either, but I'd lean toward leaving the current implementation alone.  Like @cprice404, I'd been thinking of tk's `core/main` as being the "outermost" layer of a tk application and that the general operating practice had been that slingshot messages are expected to be handled internally and not bubbled all the way back out of `core/main`.  From that perspective, I think including the extra `core/main` wrapper around the re-thrown slingshot exception is valuable in that it could help with a more consistent way of identifying a slingshot exception that should have been handled at a lower level.

KevinCorcoran:
:+1:  This seems like an improvement to me.  I think the spirit of the original implementation was to simply re-throw the original error, not "wrap and re-throw" - and that we probably *meant* to implement it this way originally (i.e., if this is Java, `throw m` is just a re-throw)

cprice404:
well then.  :)

This probably isn't major enough to be worth a ton of debate, so if it will help something in PuppetDB and some folks like @KevinCorcoran are +1, let's just merge it.  :)

PR #195, opened 2015-09-17
Refactor testutils.logging/logs-matching to match on message and level.

scotje:
Previously, logs-matching only filtered based on the log message and
given regex. The logged? assertion then checked to ensure there was
exactly 1 matching log line before considering the specified
log level. This led to the assertion failing in situations where
more than one log message matched even if one of the matches was at a
different level than the one specified in the assertion.

### DIFF
--- a/test/puppetlabs/trapperkeeper/logging_test.clj
+++ b/test/puppetlabs/trapperkeeper/logging_test.clj
@@ -46,3 +46,16 @@
       (let [lines (line-seq reader)]
         (is (= 1 (count lines)))
         (is (re-matches #".*Hi! I shouldn't get filtered\..*" (first lines)))))))
+
+(deftest test-logs-matching
+  (let [log-lines '([puppetlabs.trapperkeeper.logging-test :info nil "log message1 at info"]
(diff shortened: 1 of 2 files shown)


nwolfe:
Ah nice! I remember running into some oddities with these logging utils way back when, thanks for improving this stuff.

@scotje What do you think about a test for this behavior?

scotje:
Yeah, I could whip up some unit level tests at least for logs-matching. 

Are there any existing tests for this namespace hiding somewhere? I looked a bit and didn't see any...

nwolfe:
> Are there any existing tests for this namespace hiding somewhere? I looked a bit and didn't see any...

Yeah, most of these were just used in other tests rather getting their own test suite.  Looks like we have some logging-related tests here though, I'd probably just tack on a `deftest`: https://github.com/puppetlabs/trapperkeeper/blob/master/test/puppetlabs/trapperkeeper/logging_test.clj

scotje:
Pushed a new commit with a couple simple unit tests for logs-matching. Let me know if you think it needs more or you want me to squash.

cprice404:
+1

nwolfe:
Nice :+1:

PR #196, opened 2015-11-02
(TK-291) Handle all events in testutils.logging

rbrw:
Rework testutils.logging to tie in at the logback (Java) level, rather
than the clojure.tools.logging level, allowing us to see and manipulate
most, if not all log events.

Have the core API handle only native event objects for efficiency, but
provide a more friendly map representation via event->map.

Add a set of scoped call-with-* and with-* handlers to allow temporarily
overriding the logging appenders, level, etc.

Rework the existing with-test-logging, with-test-logging-debug, and
logged? code to use the new API, add some tests, and add some minor new,
backward-compatible features.  Using the new API does change the
fundamental semantics a bit, since logged? may see more events than it
used to (perhaps from ActiveMQ or...), but that seems unlikely to break
existing tests unless the the selectors weren't specific enough to begin
with, and is easily fixed.

Add a new with-log-suppressed-unless-notable macro that suppresses all
logging unless an event matches its notable-event? predicate, in which
case it will dump the log to *err* and provide a path to the temporary
log file.

Deprecate much of the previous code, which should for the most part be
superceded by the new calls.

cf. PDB-1995

[This began as PuppetDB 5a4789fcd3a82514a335ab59735965ef0d2a6e59]

### DIFF
--- a/test/puppetlabs/trapperkeeper/testutils/logging.clj
+++ b/test/puppetlabs/trapperkeeper/testutils/logging.clj
@@ -272,9 +272,10 @@
   [& body]
   `(let [destination# (atom [])]
      (binding [*test-log-events* destination#]
-       (with-log-level root-logger-name :trace
-         (with-logging-to-atom root-logger-name destination#
-           ~@body)))))
+       (with-redefs [pl-log/configure-logger! (fn [& _#])]
+         (with-log-level root-logger-name :trace
+                         (with-logging-to-atom root-logger-name destination#
+                                               ~@body))))))
 
 (defmacro with-test-logging-debug
   "Creates an environment for the use of the logged? test method, and
--- a/test/puppetlabs/trapperkeeper/testutils/logging.clj
+++ b/test/puppetlabs/trapperkeeper/testutils/logging.clj
@@ -1,28 +1,361 @@
 (ns puppetlabs.trapperkeeper.testutils.logging
-  (:import (java.util.regex Pattern)
-           (org.slf4j LoggerFactory Logger)
-           (ch.qos.logback.core AppenderBase))
-  (:require [clojure.tools.logging.impl :as impl]
-            [clojure.tools.logging :refer [*logger-factory*]]
-            [clojure.test :refer [assert-expr]]
-            [puppetlabs.trapperkeeper.logging :as pl-log]))
+  (:import
+   [ch.qos.logback.classic Level Logger]
+   [ch.qos.logback.classic.encoder PatternLayoutEncoder]
+   [ch.qos.logback.core Appender FileAppender]
+   [ch.qos.logback.core.spi LifeCycle]
+   (ch.qos.logback.core AppenderBase)
+   [org.slf4j LoggerFactory]
+   [java.util.regex Pattern])
+  (:require
+   [clojure.set :as set]
+   [clojure.test]
+   [clojure.tools.logging.impl :as impl]
+   [clojure.tools.logging :refer [*logger-factory*]]
+   [me.raynes.fs :as fs]
+   [puppetlabs.kitchensink.core :as kitchensink]
+   [puppetlabs.trapperkeeper.logging :as pl-log
+    :refer [root-logger-name]]
+   [schema.core :as s]))
+
+;; Note that the logging configuration is a global resource, so
+;; simultaneous calls to reset-logging, configure-logging!, etc. may
+;; interfere with many of these calls.
+
+(def ^:private keyword-levels
+  {:trace Level/TRACE
+   :debug Level/DEBUG
+   :info Level/INFO
+   :warn Level/WARN
+   :error Level/ERROR})
+
+(def ^:private level-keywords (set/map-invert keyword-levels))
+
+(def ^:private levels (set (keys keyword-levels)))
+
+(defn event->map [event]
+  "Returns {:logger name :level lvl :exception throwable :message msg}
+  for the given event.  Note that this does not convert any nil
+  messages to \"\"."
+  {:logger (.getLoggerName event)
+   :level  (level-keywords (.getLevel event))
+   :message (.getMessage event)
+   :exception (.getThrowableProxy event)})
+
+;; Perhaps make call-with-started and with-started public in
+;; kitchensink or some other namespace?
+
+(defn- call-with-started [objs f]
+  (if-not (seq objs)
+    (f)
+    (let [[obj & remainder] objs]
+      (try
+        (call-with-started remainder f)
+        (finally (.stop obj))))))
+
+(defmacro ^:private with-started
+  "Ensures that if a given name's init form executes without throwing
+  an exception, (.stop name) will be called before returning from
+  with-started.  It is the responsibility of the init form to make
+  sure the object has been started.  This macro behaves like
+  with-open, but with respect to .stop instead of .close."
+  [bindings & body]
+  (let [names (take-nth 2 bindings)
+        initializers (take-nth 2 (rest bindings))]
+    `(let [started-objs# [~@initializers]]
+       (#'puppetlabs.trapperkeeper.testutils.logging/call-with-started
+        started-objs#
+        (fn []
+          (apply (fn [~@names] ~@body) started-objs#))))))
+
+(defn- find-logger [id]
+  (LoggerFactory/getLogger (if (class? id) id (str id))))
+
+(defn call-with-log-level
+  "Sets the (logback) log level for the logger specified by logger-id
+  during the evaluation of f.  If logger-id is not a class, it will be
+  converted via str, and the level must be a clojure.tools.logging
+  key, i.e. :info, :error, etc."
+  [logger-id level f]
+  ;; Assumes use of logback (i.e. logger supports Levels).
+  (let [logger (find-logger logger-id)
+        original-level (.getLevel logger)]
+    (try
+      (.setLevel logger (level keyword-levels))
+      (f)
+      (finally (.setLevel logger original-level)))))
+
+(defmacro with-log-level
+  "Sets the (logback) log level for the logger specified by logger-id
+  during the evaluation of body.  If logger-id is not a class, it will
+  be converted via str, and the level must be a clojure.tools.logging
+  key, i.e. :info, :error, etc."
+  [logger-id level & body]
+  `(call-with-log-level ~logger-id ~level (fn [] ~@body)))
+
+(defn- log-event-listener
+  "Returns a log Appender that will call (listen event) for each log event."
+  [listen]
+  ;; No clue yet if we're supposed to start with a default name.
+  (let [name (atom (str "tk-log-listener-" (kitchensink/uuid)))]
+    (reify
+      Appender
+      (doAppend [this event] (listen event))
+      (getName [this] @name)
+      (setName [this x] (reset! name x))
+      LifeCycle
+      (start [this] true)
+      (stop [this] true))))
+
+(defn call-with-additional-log-appenders [logger-id appenders f]
+  "Adds the specified appenders to the logger specified by logger-id,
+  calls f, and then removes them.  If logger-id is not a class, it
+  will be converted via str."
+  (let [logger (find-logger logger-id)]
+    (try
+      (doseq [appender appenders]
+        (.addAppender logger appender))
+      (f)
+      (finally
+        (doseq [appender appenders]
+          (.detachAppender logger appender))))))
+
+(defmacro with-additional-log-appenders
+  "Adds the specified appenders to the logger specified by logger-id,
+  evaluates body, and then removes them.  If logger-id is not a class,
+  it will be converted via str."
+  [logger-id appenders & body]
+  `(call-with-additional-log-appenders ~logger-id ~appenders (fn [] ~@body)))
+
+(defn call-with-log-appenders [logger-id appenders f]
+  "Replaces the appenders of the logger specified by logger-id with
+  the specified appenders, calls f, and then restores the original
+  appenders.  If logger-id is not a class, it will be converted via
+  str."
+  (let [logger (find-logger logger-id)
+        original-appenders (iterator-seq (.iteratorForAppenders logger))]
+    (try
+      (doseq [appender original-appenders]
+        (.detachAppender logger appender))
+      (call-with-additional-log-appenders logger-id appenders f)
+      (finally
+        (doseq [appender original-appenders]
+          (.addAppender logger appender))))))
+
+(defmacro with-log-appenders
+  "Replaces the appenders of the logger specified by logger-id with
+  the specified appenders, evaluates body, and then restores the
+  original appenders.  If logger-id is not a class, it will be
+  converted via str."
+  [logger-id appenders & body]
+  `(call-with-log-appenders ~logger-id ~appenders (fn [] ~@body)))
+
+(defn call-with-additional-log-event-listeners
+  "For each listen in listens, calls (listen event) for each logger-id
+  event produced during the evaluation of f.  If logger-id is not a
+  class, it will be converted via str."
+  [logger-id listens f]
+  (letfn [(set-up [listens listeners]
+            (if-not (seq listens)
+              (call-with-additional-log-appenders logger-id listeners f)
+              (let [[listen & remainder] listens]
+                (with-started [listener (doto (log-event-listener listen)
+                                          .start)]
+                  (set-up remainder (conj listeners listener))))))]
+    (set-up listens [])))
+
+(defmacro with-additional-log-event-listeners
+  "For each listen in listens, calls (listen event) for each logger-id
+  event produced during the evaluation of body.  If logger-id is not a
+  class, it will be converted via str."
+  [logger-id listens & body]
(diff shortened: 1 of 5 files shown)


cprice404:
@rbrw thanks a bunch for filing this.  Our logging testutils have been driving me nuts lately but I haven't had time to do anything.

Probably won't have bandwidth to review this for a bit due to Ankeny time pressures, but will review ASAP.

test/puppetlabs/trapperkeeper/testutils/logging_test.clj
```diff
@@ -0,0 +1,150 @@
...
+   [clojure.test :refer :all]
+   [clojure.tools.logging :as log]
+   [puppetlabs.kitchensink.core :as kitchensink]
+   [puppetlabs.trapperkeeper.logging :refer [reset-logging root-logger-name]]
+   [puppetlabs.trapperkeeper.testutils.logging :as tgt :refer [event->map]])
+  (:import
+   [org.slf4j Logger LoggerFactory]))
+
+;; Without this, "lein test NAMESPACE" and :only invocations may fail.
+(use-fixtures :once (fn [f] (reset-logging) (f)))
```
dankreek: I'm curious about why this is needed.

test/puppetlabs/trapperkeeper/testutils/logging.clj
```diff
@@ -1,28 +1,321 @@
...
+
+(def ^:private level-keywords (set/map-invert keyword-levels))
+
+(def ^:private levels (set (keys keyword-levels)))
+
+(defn event->map [event]
+  "Returns {:logger name :level lvl :exception throwable :message msg}
+  for the given event.  Note that this does not convert any nil
+  messages to \"\"."
+  (merge
```
dankreek: Merging a single map?

rbrw: Ahh thanks, that's vestigial; I'll fix it.

test/puppetlabs/trapperkeeper/testutils/logging_test.clj
rbrw: I'd assumed it was because in the `lein test' case, some other namespace (that runs first) was initializing logging, but for something like `lein test puppetlabs.trapperkeeper.testutils.logging-test`, we need to do it manually (via the fixture).

dankreek:
So I've attempted a functional review using this branch as a checkout while running the Puppet sever tests. First thing I've noticed is that the logout output is being dumped to the console now, where as it wasn't before.

test/puppetlabs/trapperkeeper/testutils/logging.clj
```diff
@@ -1,28 +1,321 @@
...
+
+(defmacro ^:private with-started
+  "Ensures that if a given name's init form executes without throwing
+  an exception, (.stop name) will be called before returning from
+  with-started.  It is the responsibility of the init form to make
+  sure the object has been started.  This macro behaves like
+  with-open, but with respect to .stop instead of .close."
+  [bindings & body]
+  ;; We call out to a function because Clojure currently chokes on a
+  ;; recursive private macro.
```
cprice404: I think I'm going to start a band called "recursive private macro".

cprice404: FYI I am actually trying to review this, not just leave hilarious comments.  Just keep getting interrupted.

test/puppetlabs/trapperkeeper/testutils/logging_test.clj
```diff
@@ -0,0 +1,150 @@
...
+          (log/info "wlta-test"))
+        (is (not-any? #(= expected %) (map event->map @log)))))
+    (let [log (atom [])]
+      (tgt/with-log-level root-logger-name :info
+        (tgt/with-logging-to-atom root-logger-name log
+          (log/info "wlta-test"))
+        (is (some #(= expected %) (map event->map @log)))))))
+
+(def find-logger #'puppetlabs.trapperkeeper.testutils.logging/find-logger)
+(def log-event-listener #'puppetlabs.trapperkeeper.testutils.logging/log-event-listener)
```
cprice404: Why `def`s instead of `require`s for these?

test/puppetlabs/trapperkeeper/testutils/logging_test.clj
```diff
@@ -0,0 +1,150 @@
...
+                                  (and (= :trace (:level e))
+                                       (= "baz" (:message e))))]]]
+      (tgt/with-test-logging
+        (log/trace item)
+        (is (logged? test)))
+      (tgt/with-test-logging
+        (log/trace "hapax legomenon")
+        (is (not (tgt/logged? test))))))
+  (testing "level matches"
+    (doseq [level @#'puppetlabs.trapperkeeper.testutils.logging/levels]
```
cprice404: Why do you need to deref the symbol here instead of just referring to `levels` directly?

test/puppetlabs/trapperkeeper/testutils/logging.clj
```diff
@@ -1,28 +1,321 @@
...
+  `(with-log-event-listener
+     ~logger-id
+     (fn [event#] (swap! ~destination conj event#))
+     ~@body))
+
+(defn- suppressing-file-appender
+  [log-path]
+  (let [pattern "%-4relative [%thread] %-5level %logger{35} - %msg%n"
+        context (LoggerFactory/getILoggerFactory)]
+    (doto (FileAppender.)
```
cprice404: Is it necessary to write to a file for this one?  Or just an easier implementation path?  No big deal either way, just curious.

test/puppetlabs/trapperkeeper/testutils/logging.clj
```diff
@@ -1,28 +1,321 @@
...
+          (fs/delete log-path))))))
+
+(defmacro with-log-suppressed-unless-notable
+  "Executes the body with all logging suppressed.  If a notable error
+  is logged, dumps the full log to *err* along with its path.  Assumes
+  that the logging level is already set as desired.  This may not work
+  correctly if the system logback config is altered during the
+  execution of the body."
+  [notable-event? & body]
+  `(#'puppetlabs.trapperkeeper.testutils.logging/call-with-log-suppressed-unless-notable
```
cprice404: Is the symbol use here is necessary because it's a macro?

test/puppetlabs/trapperkeeper/testutils/logging_test.clj
cprice404: ah, I see, it's because those are private.  Is this the pattern you guys typically use instead of just making the functions public?

test/puppetlabs/trapperkeeper/testutils/logging.clj
```diff
@@ -1,28 +1,321 @@
...
+            (print (slurp log-path))
+            (println "From error log:" log-path))
+          (fs/delete log-path))))))
+
+(defmacro with-log-suppressed-unless-notable
+  "Executes the body with all logging suppressed.  If a notable error
+  is logged, dumps the full log to *err* along with its path.  Assumes
+  that the logging level is already set as desired.  This may not work
+  correctly if the system logback config is altered during the
+  execution of the body."
```
cprice404: Might be worth clarifying this docstring to clarify how an event qualifies as `notable`, and that `notable-event?` is a predicate function.

test/puppetlabs/trapperkeeper/testutils/logging.clj
```diff
@@ -1,28 +1,321 @@
...
 (def ^{:doc "A dynamic var that is bound to an atom containing all of the log entries
              that have occurred during a test, when using `with-test-logging`."
-       :dynamic true}
+       :dynamic true
+       :deprecated "1.1.2"}
   *test-logs*
   nil)
 
-(def legal-levels #{nil :trace :debug :info :warn :error :fatal})
+(def ^{:deprecated "1.1.2"} legal-levels
```
cprice404: Does clojure provide any guidance on deprecation metadata?  Is there a field where you can specify what the replacement is for the deprecated thing?  And does clojure surface any kind of warning about using deprecated things?

Also I think we might consider sending an email to the dev list asking if anyone would find it horribly problematic for us to just delete these rather than deprecate them... though I guess it's theoretically possible that some OSS users are using them.

cprice404:
I had minor random questions and comments but overall this looks great to me... going to spend some time playing with it with Puppet Server.

cprice404:
Testing a bit just in the TK project itself... some good news and some less good.

Good news:

This PR seems to also resolve TK-269, which is AWESOME.

Less good:

Even in the TK repo itself, this PR causes the tests to log a bunch of stuff that they weren't logging before to the console.  Same as what @dankreek was pointing out, almost certainly; but we sometimes use `with-test-logging` just to silence a noisy test so that it doesn't make the lein output hard to sift through.  I'm sure there's an easy fix to get that working with this PR though.

cprice404:
@rbrw I have been poking around with this code, and besides my rando questions/comments above, I threw together a couple of commits for consideration.

This one:

https://github.com/cprice404/trapperkeeper/commit/d58a6edf163027bf5836899f264f2acb0be359fc

just adds some tests that validate that this fixes TK-269.

This commit:

https://github.com/cprice404/trapperkeeper/commit/f7737e11649c9e793a030b2955945d9ebda005d7

changes the behavior of `with-test-logging` such that it removes the existing loggers before running the test, which surpresses most of the log output.  This makes it match the previous behavior more closely.  I haven't quite verified that it *exactly* matches the old behavior, but for the puppet server tests, without this change, they are logging a very large volume of stuff that they didn't used to.  With this change, they're logging significantly less, but maybe still just a bit more than what they did before.  We can dig into that a bit further and at least verify whether the behavior changes one way or another, but, thought I'd at least throw the commit up there for you to look at and consider.

Thanks!

cprice404:
hrm.  Even with those commits I mentioned above, there are some places in the puppetserver tests where we're using `with-test-logging`, and some messages at level `WARN` or above are being logged during the test runs, where they weren't previously.  Trying to determine why.

cprice404:
So it looks like the issue is that basically any test that uses 'with-app-with-config` (which we use a lot) will end up calling the "for realz" TK `configure-logging` function *after* the `with-test-logging` has done its manipulation of the root logger.  The TK function will then call `reset` on the main logback logging context, which I assume effectively blows away the changes.  Not quite sure what to do about that.

If we can't figure out a way to solve this in the logging test utils, then I still think the changes in this PR are valuable enough that I wouldn't have a problem tweaking the offending tests in other repos in order to get this patch merged, but, we probably at least need to figure out what the changes to said tests would look like.

cprice404:
OK, adding this commit:

https://github.com/cprice404/trapperkeeper/commit/1a83d4f18818e26ec75db658482d2a62cba67d25

solves it.  It does a `with-redef` on the "real" `configure-logger!` function so that, once inside of a `with-test-logging` block, the other test utilities don't end up blowing away the logger changes.  With these commits applied to your branch, I am now getting the same output from the puppet server tests that I was before.

@rbrw no rush, I'm sure this isn't urgent priority for you, but I'm a big +1 on getting this stuff in so whenever you have time to respond and share your thoughts, I'm happy to help however I can.

test/puppetlabs/trapperkeeper/testutils/logging_test.clj
rbrw: @cprice404 if "we" might mean just me at the moment, then yes?  It's just a way I've used to test things that are private and/or use them in tests.

cprice404: Since I never wrote Clojure before PL I don't always feel like I know what the best practices are, but, my knee-jerk reaction to this pattern is concern that it would prevent things like automatic refactorings in dev environments from catching these references.  So we've just erred on the side of not making things private.  If we consider them internal-only then we might do something like suffix the function name with an asterisk or move it to a separate namespace called `internal` just to avoid confusion about what should be considered public API.  Personally I prefer those approaches, but I don't feel strongly enough about it to where if some folks who had more Clojure experience than myself said "this is a normal way to do things" that I'd resist too heavily.

rbrw: Well, I'm certainly not the expert, but I don't think there's any consensus on this front.  Even within our team, I believe opinions vary.  Personally, I mostly use private as a signal that "this thing isn't supported for external use".   Then, when I'm working in a namespace, aside from explicit circumventions like this (presumably only in testing code), I know that I should have substantial freedom to rework/remove anything private because no one else could have been depending on it (without an intentional sidestep).  And of course that's even more valuable when working on shared libraries.  In any case, I don't have a categorical opinion on the subject, just explaining my current approach.

(Mentioning @senior in case he wants to weigh in, since we've also discussed this from time to time.)

test/puppetlabs/trapperkeeper/testutils/logging_test.clj
rbrw: I believe because it's a private var (levels and level-keywords are just internal "caches").  And removing it produces "java.lang.IllegalArgumentException: Don't know how to create ISeq from: clojure.lang.Var".

test/puppetlabs/trapperkeeper/testutils/logging.clj
rbrw: Because it's private, and will need to be called from within the with-log-suppressed-unless-notable caller's namespace, after macro expansion.

rbrw: Though actually, I should change this, since I think that with-foo macros should often (normally) provide a parallel, supporting call-with-foo function.  I'll make the call-with- public.

test/puppetlabs/trapperkeeper/testutils/logging_test.clj
cprice404: OK.  I agree with the goals but we've just been using different tools to achieve them (pragmas in the code, internal namespaces, etc.).  It's not a huge deal, though.

test/puppetlabs/trapperkeeper/testutils/logging.clj
rbrw: Right - I just renamed it to "pred" to match the existing Clojure doc convention, and reworked the docs to hopefully be a bit clearer.

test/puppetlabs/trapperkeeper/testutils/logging.clj
rbrw: Don't think it's required per-se.  We could change it, though I suppose that could mean a lot more data in RAM given any heavy logging, and it'd leave no trace after a hard crash.

rbrw:
@cprice404 ok, I *think* that update may address the issues raised so far.

rbrw:
@cprice404 and it includes what I thought were the pertinent patches you settled on.

test/puppetlabs/trapperkeeper/testutils/logging.clj
senior: This metadata is used quite a bit in [clojure.core](https://github.com/clojure/clojure/blob/bdc752a7fefff5e63e0847836ae5e6d95f971c37/src/clj/clojure/core.clj#L4924). There is nothing special in the language, but autodoc does handle it [correctly](https://clojure.github.io/clojure/clojure.core-api.html#clojure.core/add-classpath).  I've not used [deprecated](https://github.com/alexander-yakushev/defprecated) but that does add more rigor along with warnings.

Seems like deprecating it and removing it on the next major is the right approach here. We publish and encourage the usage of test-jars, seems like it should have the same semver properties.

I wouldn't be a fan of removing it without a semver bump. Seems like it should follow the rest

cprice404: alrighty then

cprice404:
@rbrw I think we also need this commit:

https://github.com/cprice404/trapperkeeper/commit/f7737e11649c9e793a030b2955945d9ebda005d7

Other than that I'm +1.

test/puppetlabs/trapperkeeper/testutils/logging.clj
cprice404: ok

cprice404:
+1 this works for me now.

@pcarlisle do you or anyone else in your part of the world want to review this before it gets merged?

pcarlisle:
I think we're good.

cprice404:
@KevinCorcoran @camlow325 @nwolfe any of you guys?  I'm ready to push the button on this unless one of you is interested in reviewing first.

nwolfe:
:see_no_evil:

KevinCorcoran:
I'm interested in looking at this but don't hold up merge on my account.  I've been :rage: at these utilities for a long time so I'm happy that ye put in the work to improve them.  Doubt I'll have any significant time to dive into this any time soon.

camlow325:
Same here.  I did a cursory pass of this a while back and didn't see anything concerning.  Looks like @cprice404 has done a much more thorough review and testing so I trust his judgment on this being merged if he's good with it.  Will be cool to see if this allows us to get rid of some of the remaining unwanted log messages we see from Jetty in stdout during lein test runs using trapperkeeper-webservices-jetty9 in the stack.

PR #198, opened 2015-12-15
(TK-202) Add support for restart and HUP handling

haus:
This PR adds a restart function for TK apps that allows trapperkeeper to cleanly handle SIGHUP by calling restart on all of the registered apps.

### DIFF
--- a/src/puppetlabs/trapperkeeper/app.clj
+++ b/src/puppetlabs/trapperkeeper/app.clj
@@ -12,4 +12,5 @@
                                  "occurred, return the input parameter."))
   (init [this] "Initialize the services")
   (start [this] "Start the services")
-  (stop [this] "Stop the services"))
+  (stop [this] "Stop the services")
+  (restart [this] "Stop and restart the services"))
--- a/src/puppetlabs/trapperkeeper/core.clj
+++ b/src/puppetlabs/trapperkeeper/core.clj
@@ -32,13 +32,14 @@
   (`init`, `start`, `stop`) as you see fit;  if you'd like to have the trapperkeeper
   framework block the main thread to wait for a shutdown event, call `init`,
   `start`, and then `run-app`."
-  [services config-data-fn]
+  [services config-data]
   {:pre  [(sequential? services)
           (every? #(satisfies? services/ServiceDefinition %) services)
-          (ifn? config-data-fn)]
+          (ifn? config-data)]
    :post [(satisfies? app/TrapperkeeperApp %)]}
-  (config/initialize-logging! (config-data-fn))
-  (internal/build-app* services config-data-fn (promise)))
+  (let [config-data-fn (if (map? config-data) (constantly config-data) config-data)]
+    (config/initialize-logging! (config-data-fn))
+    (internal/build-app* services config-data-fn (promise))))
 
 (defn boot-services-with-cli-data
   "Given a list of ServiceDefinitions and a map containing parsed cli data, create
--- a/src/puppetlabs/trapperkeeper/core.clj
+++ b/src/puppetlabs/trapperkeeper/core.clj
@@ -114,13 +114,6 @@
         (bootstrap/parse-bootstrap-config!)
         (internal/boot-services* config-data))))
 
-
-;; This variable is used to hold a reference to the "main" trapperkeeper
-;; application instance if you use trapperkeeper's "main" function to
-;; launch your application.  This allows you to get a reference to the
-;; app in a remote nREPL session if you are using the nrepl service.
-(def main-app nil)
-
 (defn run-app
   "Given a bootstrapped TrapperKeeper app, let the application run until shut down,
   which may be triggered by one of several different ways. In all cases, services
@@ -144,7 +137,7 @@
   (let [app (boot-with-cli-data cli-data)]
     ;; This line populates the `main-app` variable with the TrapperkeeperApp
     ;; instance.  This allows it to be referenced in a remote nREPL session.
-    (alter-var-root #'main-app (fn [_] app))
+    (swap! internal/tk-apps conj app)
     (run-app app)))
 
 (defn main
--- a/src/puppetlabs/trapperkeeper/core.clj
+++ b/src/puppetlabs/trapperkeeper/core.clj
@@ -32,13 +32,13 @@
   (`init`, `start`, `stop`) as you see fit;  if you'd like to have the trapperkeeper
   framework block the main thread to wait for a shutdown event, call `init`,
   `start`, and then `run-app`."
-  [services config-data]
+  [services config-data-fn]
   {:pre  [(sequential? services)
           (every? #(satisfies? services/ServiceDefinition %) services)
-          (map? config-data)]
+          (ifn? config-data-fn)]
    :post [(satisfies? app/TrapperkeeperApp %)]}
-  (config/initialize-logging! config-data)
-  (internal/build-app* services config-data (promise)))
+  (config/initialize-logging! (config-data-fn))
+  (internal/build-app* services config-data-fn (promise)))
 
 (defn boot-services-with-cli-data
   "Given a list of ServiceDefinitions and a map containing parsed cli data, create
@@ -54,15 +54,33 @@
           (every? #(satisfies? services/ServiceDefinition %) services)
           (map? cli-data)]
    :post [(satisfies? app/TrapperkeeperApp %)]}
-  (let [config-data (config/parse-config-data cli-data)]
-    (config/initialize-logging! config-data)
-    (internal/boot-services* services config-data)))
+  (let [config-data-fn #(config/parse-config-data cli-data)]
+    (config/initialize-logging! (config-data-fn))
+    (internal/boot-services* services config-data-fn)))
+
+(defn boot-services-with-config-fn
+  "Given a list of ServiceDefinitions and a function that returns a
+  map containing parsed cli data, create and boot a trapperkeeper app.  This
+  function can be used if you prefer to do your own CLI parsing and loading
+  ServiceDefinitions; it circumvents the normal trapperkeeper `bootstrap.cfg`
+  boot process, but still allows trapperkeeper to handle the parsing of your
+  service configuration data.
+
+  Returns a TrapperkeeperApp instance.  Call `run-app` on it if you'd like to
+  block the main thread to wait for a shutdown event."
+  [services config-data-fn]
+  {:pre  [(sequential? services)
+          (every? #(satisfies? services/ServiceDefinition %) services)
+          (ifn? config-data-fn)]
+   :post [(satisfies? app/TrapperkeeperApp %)]}
+  (config/initialize-logging! (config-data-fn))
+  (internal/boot-services* services config-data-fn))
 
 (defn boot-services-with-config
-  "Given a list of ServiceDefinitions and a map containing parsed cli data, create
-  and boot a trapperkeeper app.  This function can be used if you prefer to
-  do your own CLI parsing and loading ServiceDefinitions; it circumvents
-  the normal trapperkeeper `bootstrap.cfg` boot process, but still allows
+  "Given a list of ServiceDefinitions and a map containing parsed cli data,
+  create and boot a trapperkeeper app.  This function can be used if you prefer
+  to do your own CLI parsing and loading ServiceDefinitions; it circumvents the
+  normal trapperkeeper `bootstrap.cfg` boot process, but still allows
   trapperkeeper to handle the parsing of your service configuration data.
 
   Returns a TrapperkeeperApp instance.  Call `run-app` on it if you'd like to
@@ -72,8 +90,7 @@
           (every? #(satisfies? services/ServiceDefinition %) services)
           (map? config-data)]
    :post [(satisfies? app/TrapperkeeperApp %)]}
-  (config/initialize-logging! config-data)
-  (internal/boot-services* services config-data))
+  (boot-services-with-config-fn services (constantly config-data)))
 
 (defn boot-with-cli-data
   "Create and boot a trapperkeeper application.  This is accomplished by reading a
@@ -106,13 +123,13 @@
   ;; 2. initialize logging
   ;; 3. initialize plugin system
   ;; 4. bootstrap rest of framework
-  (let [config-data (config/parse-config-data cli-data)]
-    (config/initialize-logging! config-data)
+  (let [config-data-fn #(config/parse-config-data cli-data)]
+    (config/initialize-logging! (config-data-fn))
     (plugins/add-plugin-jars-to-classpath! (cli-data :plugins))
     (-> cli-data
         (bootstrap/find-bootstrap-config)
         (bootstrap/parse-bootstrap-config!)
-        (internal/boot-services* config-data))))
+        (internal/boot-services* config-data-fn))))
 
 (defn run-app
   "Given a bootstrapped TrapperKeeper app, let the application run until shut down,
--- a/src/puppetlabs/trapperkeeper/core.clj
+++ b/src/puppetlabs/trapperkeeper/core.clj
@@ -138,6 +138,7 @@
     ;; This adds the TrapperkeeperApp instance to the tk-apps list, so that
     ;; it can be referenced in a remote nREPL session, etc.
     (swap! internal/tk-apps conj app)
+    (internal/register-sighup-handler)
     (run-app app)
     (swap! internal/tk-apps (partial remove #{app}))))
 
--- a/src/puppetlabs/trapperkeeper/core.clj
+++ b/src/puppetlabs/trapperkeeper/core.clj
@@ -135,10 +135,11 @@
   [cli-data]
   {:pre [(map? cli-data)]}
   (let [app (boot-with-cli-data cli-data)]
-    ;; This line populates the `main-app` variable with the TrapperkeeperApp
-    ;; instance.  This allows it to be referenced in a remote nREPL session.
+    ;; This adds the TrapperkeeperApp instance to the tk-apps list, so that
+    ;; it can be referenced in a remote nREPL session, etc.
     (swap! internal/tk-apps conj app)
-    (run-app app)))
+    (run-app app)
+    (swap! internal/tk-apps (partial remove #{app}))))
 
 (defn main
   "Launches the trapperkeeper framework. This function blocks until
--- a/src/puppetlabs/trapperkeeper/internal.clj
+++ b/src/puppetlabs/trapperkeeper/internal.clj
@@ -478,6 +478,7 @@
         this)
       (a/restart [this]
         (run-lifecycle-fns app-context s/stop "stop" (reverse ordered-services))
+        (doseq [svc-id (keys services-by-id)] (swap! app-context assoc svc-id {}))
         (run-lifecycle-fns app-context s/init "init" ordered-services)
         (run-lifecycle-fns app-context s/start "start" ordered-services)))))
 
--- a/src/puppetlabs/trapperkeeper/internal.clj
+++ b/src/puppetlabs/trapperkeeper/internal.clj
@@ -477,10 +477,14 @@
         (shutdown! app-context)
         this)
       (a/restart [this]
-        (run-lifecycle-fns app-context s/stop "stop" (reverse ordered-services))
-        (doseq [svc-id (keys services-by-id)] (swap! app-context assoc svc-id {}))
-        (run-lifecycle-fns app-context s/init "init" ordered-services)
-        (run-lifecycle-fns app-context s/start "start" ordered-services)))))
+        (try
+          (run-lifecycle-fns app-context s/stop "stop" (reverse ordered-services))
+          (doseq [svc-id (keys services-by-id)] (swap! app-context assoc svc-id {}))
+          (run-lifecycle-fns app-context s/init "init" ordered-services)
+          (run-lifecycle-fns app-context s/start "start" ordered-services)
+          (catch Throwable t
+            (deliver shutdown-reason-promise {:cause :service-error
+                                              :error t})))))))
 
 (defn boot-services*
   "Given the services to run and the map of configuration data, create the
--- a/src/puppetlabs/trapperkeeper/internal.clj
+++ b/src/puppetlabs/trapperkeeper/internal.clj
@@ -8,6 +8,8 @@
             [puppetlabs.trapperkeeper.services :as s]
             [puppetlabs.kitchensink.core :as ks]))
 
+(def tk-apps (atom []))
+
 (defn service-graph?
   "Predicate that tests whether or not the argument is a valid trapperkeeper
   service graph."
--- a/src/puppetlabs/trapperkeeper/internal.clj
+++ b/src/puppetlabs/trapperkeeper/internal.clj
@@ -430,10 +430,10 @@
 (defn build-app*
   "Given a list of services and a map of configuration data, build an instance
   of a TrapperkeeperApp.  Services are not yet initialized or started."
-  [services config-data shutdown-reason-promise]
+  [services config-data-fn shutdown-reason-promise]
   {:pre  [(sequential? services)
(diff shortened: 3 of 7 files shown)


igalic:
https://github.com/travis-ci/travis-ci/issues/5227

cprice404:
@igalic thanks for the heads up.  That's disconcerting :/

test/puppetlabs/trapperkeeper/services_test.clj
```diff
@@ -49,56 +51,137 @@
...
+       (stop [this context] (lc-fn context :stop-service2))
+       (service2-fn [this] (lc-fn nil :service2-fn)))
+
+     (service
+       Service3
+       [[:Service2 service2-fn]]
+       (init [this context] (lc-fn context :init-service3))
+       (start [this context] (lc-fn context :start-service3))
+       (stop [this context] (lc-fn context :stop-service3))
+       (service3-fn [this] (lc-fn nil :service3-fn)))]))
```
cprice404: I might suggest moving these down so that they're just above the tests that use them.

cprice404:
+1 other than the one minor nitpick.

Probably also worth doing a `lein install` and testing this with Puppet Server to make sure we didn't accidentally break any backward compat.

cprice404:
@pcarlisle do you want to review this?  @haus added a test that actually exercises the signal.

pcarlisle:
cool, looks fine to me

test/puppetlabs/trapperkeeper/services_test.clj
```diff
@@ -49,56 +51,137 @@
...
+          services (create-lifecycle-services call-seq)]
+      (with-app-with-empty-config app
+        services
+        (is (= [:init-service1 :init-service2 :init-service3
+                :start-service1 :start-service2 :start-service3]
+               @call-seq))
+        (internal/register-sighup-handler [app])
+        (beckon/raise! "HUP")
+        (let [start (System/currentTimeMillis)]
+          (while (or (not= (count @call-seq) 15)
```
haus: This loop logic is inverted. I'll get a fix out this afternoon.

haus:
supposedly travis is no longer overflowing its heap, so i kicked the tests. :8ball:

src/puppetlabs/trapperkeeper/internal.clj
```diff
@@ -448,20 +475,24 @@
         this)
       (a/stop [this]
         (shutdown! app-context)
-        this))))
+        this)
+      (a/restart [this]
+        (run-lifecycle-fns app-context s/stop "stop" (reverse ordered-services))
```
camlow325: What should happen if one of these fail - e.g., if an unhandled exception is thrown up the stack?  Should the process that the apps are running in be shut down?

haus: It makes sense to me to shut the app process down in that case, but maybe @cprice404 has thoughts here.

camlow325: I think in `shutdown` we'll catch an Exception thrown from an individual service's stop function, log it, and continue calling the other services so they still have a chance to stop?  Should we do that here too?  Maybe we should share some of this logic between the restart and `shutdown!`?

camlow325:
Not that I think it's a big deal or even desirable, true statement that we won't reparse the bootstrap.cfg - potentially starting up different services than were present at initial start - after a SIGHUP, right?

test/puppetlabs/trapperkeeper/services_test.clj
```diff
@@ -49,56 +51,137 @@
...
+          services (create-lifecycle-services call-seq)]
+      (with-app-with-empty-config app
+        services
+        (is (= [:init-service1 :init-service2 :init-service3
+                :start-service1 :start-service2 :start-service3]
+               @call-seq)))
       (is (= [:init-service1 :init-service2 :init-service3
-              :start-service1 :start-service2 :start-service3]
+              :start-service1 :start-service2 :start-service3
+              :stop-service3 :stop-service2 :stop-service1]
```
camlow325: Nice.

test/puppetlabs/trapperkeeper/services_test.clj
```diff
@@ -49,56 +51,137 @@
...
+        services
+        (is (= [:init-service1 :init-service2 :init-service3
+                :start-service1 :start-service2 :start-service3]
+               @call-seq))
+        (internal/register-sighup-handler [app])
+        (beckon/raise! "HUP")
+        (let [start (System/currentTimeMillis)]
+          (while (and (not= (count @call-seq) 15)
+                     (< (- (System/currentTimeMillis) start) 1000))
+            (Thread/yield)))
```
camlow325: Is this better than a short sleep?  Doesn't matter that much, I guess.

haus: I don't think we need this for the signal handler, but it seemed slightly better than a sleep.

camlow325:
In earlier descriptions about this feature, I had the impression that it wasn't going to be possible for a service to store state in its context and then be able to get that state back after a restart.  But it looks like it does actually preserve the service specific context across the restart.  Is that intentional?

I suppose that might be nice for some services that want to conditionally skip re-initializing something if they can preserve it across a restart but probably worth documenting that as such so that services are prepared for that.  For example, it might be bad for the service to end up with a "dirty" context that is a combination of some new stuff the service wanted to add to the context after the restart but wasn't aware of old unintended stuff was implicitly preserved on the context from the previous service startup.

cprice404:
@camlow325 good catch, no, that was not intentional, we should most definitely be clearing state.

camlow325:
Overall, looks really good and seemed to do what I expected when testing it.

I left some questions about expected error handling - like should the process get shut down if the restart fails - and whether or not data is supposed to be preserved in the service context across a restart.

@cprice404 and I had also talked about the concern I had about the restart being executed from the signal handler's thread while concurrent activity might be happening on the Trapperkeeper main thread - and specifically how it might be better to make assurances that lifecycle functions only ever be executed from the same thread to avoid potential issues.  That led to https://tickets.puppetlabs.com/browse/TK-300 being filed.  I'm good with letting that question be sorted out there.

src/puppetlabs/trapperkeeper/internal.clj
haus: I've just pushed out a new commit that wraps the restart in a try-catch to achieve this effect.

haus:
@camlow325 I pushed some new commits to ensure context is being cleared on a restart.

test/puppetlabs/trapperkeeper/services_test.clj
```diff
@@ -49,56 +55,180 @@
...
+                                                                EmptyService
+                                                                []
+                                                                (stop [this context] (throw (IllegalStateException. "Exploding Service")))))]
+      (try
+        (ks-testutils/with-no-jvm-shutdown-hooks
+         (let [app (internal/throw-app-error-if-exists!
+                    (bootstrap-services-with-empty-config services))]
+           ; We can't use the with-app-with-empty-config macro because we don't want to use its implicit tk-app/stop call
+           ; We're asserting that the stop will happen because of the exception. So instead, we use the tk/run-app here to
+           ; block on the app until the restart is called an explodes in an exception.
```
KevinCorcoran: These lines are really long - we typically break them at ~80 characters (in Clojure).

KevinCorcoran:
I took a quick peek.  :see_no_evil:

test/puppetlabs/trapperkeeper/services_test.clj
haus: cool. will fix up.

haus:
okay travis. let's try this again.

haus:
check out that green from travis

cprice404:
We may need to be careful about the timing of merging this since it will have pretty major implications on users of TK.  We should perhaps consider reviewing and merging #199 first, doing a release that contains that, and then merging this afterward.  I will review #199 ASAP.

haus:
@cprice404 sounds reasonable to me.

test/puppetlabs/trapperkeeper/services_test.clj
```diff
@@ -49,56 +55,182 @@
...
+(deftest test-exception-during-restart
+  (testing "restart should halt if an exception is raised"
+    (let [call-seq (atom [])
+          services (conj (create-lifecycle-services call-seq) (service
+                                                                EmptyService
+                                                                []
+                                                                (stop [this context] (throw (IllegalStateException. "Exploding Service")))))]
+      (try
+        (ks-testutils/with-no-jvm-shutdown-hooks
+         (let [app (internal/throw-app-error-if-exists!
```
camlow325: I think you could combine this let block with the next one.

test/puppetlabs/trapperkeeper/services_test.clj
```diff
@@ -49,56 +55,182 @@
...
+      (is (= {} @test-context)))))
+
+(deftest test-exception-during-restart
+  (testing "restart should halt if an exception is raised"
+    (let [call-seq (atom [])
+          services (conj (create-lifecycle-services call-seq) (service
+                                                                EmptyService
+                                                                []
+                                                                (stop [this context] (throw (IllegalStateException. "Exploding Service")))))]
+      (try
```
camlow325: try block with no associated catch or finally block?

test/puppetlabs/trapperkeeper/services_test.clj
haus: Yup. seems to work. pushed out a commit with that change.

test/puppetlabs/trapperkeeper/services_test.clj
haus: Probably leftover from a previous iteration. removed in the most recent commit. thanks!

src/puppetlabs/trapperkeeper/internal.clj
```diff
@@ -448,20 +475,29 @@
         this)
       (a/stop [this]
         (shutdown! app-context)
-        this))))
+        this)
+      (a/restart [this]
+        (try
+          (run-lifecycle-fns app-context s/stop "stop" (reverse ordered-services))
```
camlow325: One thing that's different about this vs. what happens in `shutdown!` for an app stop is that if an exception bubbles from the `stop` lifecycle function called for one service, that exception is swallowed (but with an error message written to the log) but `stop` is still called on all of the other dependent services so that those services still have a chance to clean up their resources.  This implementation, however, would appear to cause the stop to abort after the first service's `stop` were to throw an exception.

I'm not sure if the `shutdown!` implementation is the best for all cases - e.g., if an aborted `stop` on one service were to cause the downstream `stop` calls on other services to behave abnormally for some reason.  I think it would be best, though, for a shutdown triggered from `stop` on the restart to behave the same as it would if triggered from app/stop.  So maybe the 'swallow the exception and move on to stopping the next service' approach is best for now.  @cprice404 - what do you think?

haus: I think the intention was to still call stop on the other services. My test probably didn't capture that goal as well as it should have.

haus: I'll look at making that change.

src/puppetlabs/trapperkeeper/internal.clj
```diff
@@ -448,20 +475,29 @@
...
+          (doseq [svc-id (keys services-by-id)] (swap! app-context assoc svc-id {}))
+          (run-lifecycle-fns app-context s/init "init" ordered-services)
+          (run-lifecycle-fns app-context s/start "start" ordered-services)
+          (catch Throwable t
+            (deliver shutdown-reason-promise {:cause :service-error
```
camlow325: Hmmm, one of the behaviors we might see with this as well is that `stop` could end up being called twice on each service - once from within `restart` but then again after the shutdown promise is realized and it calls back into app/stop.

@cprice404 - do you think this will be a big problem?  Not sure how to work around this without making more fundamental changes to have the restart sequence worked into the lifecycle management on the Trapperkeeper main thread or the like...

camlow325:
:+1: with just the remaining questions about stop/shutdown processing in the event of a restart failure.

src/puppetlabs/trapperkeeper/internal.clj
haus: @camlow325 this implementation actually does perform all stops, even if one errors. `run-lifecycle-fns` calls a doseq and catches exceptions, logging them as errors as it goes (almost identially to the body of `shutdown!`. I can push out an updated test still, because i don't think my test captured this well.

camlow325: > @camlow325 this implementation actually does perform all stops, even if one errors.  

`run-lifecycle-fns` will catch an exception but then will re-throw it, causing the rest of the `stop` calls to not be made in this case:

https://github.com/puppetlabs/trapperkeeper/blob/1.1.3/src/puppetlabs/trapperkeeper/internal.clj#L186

So I think the first time an exception is encountered, control will fall through to catch block, where the `shutdown-promise` is delivered.

Although, as I mentioned in a separate comment, I think `stop` will end up getting called a second time as part of the shutdown logic on the Trapperkeeper main thread after the shutdown promise is handled.  So maybe my comment here is moot.  I'm actually a little more concerned about the double-stop call case.  Essentially, I think it will mean that all service implementation `stop` methods need to be resilient to being called multiple times without an init/start in between.  Maybe that's not the end of the world either, though.  Still curious what @cprice404 thinks about this.

haus: oh right. i missed the throw there.

haus: In that case i'm not sure what the right path is. I'll wait for @cprice404 to weigh in before I do anything.

cprice404: @camlow325 I think the normal TK shutdown logic is why we coded it this way... relying on TK to call the remaining `stop` fns.  As for double-stop, I'd definitely prefer to *not* impose a requirement that services be resilient to stop being called twice, but I'm not sure I see an easy way to achieve that, so, not sure it's worth the ROI?  Happy to discuss that further, though...

We kind of have a similar issue already, in that if an error occurs during startup, then the `stop` functions may be called w/o `init` or `start` being executed successfully, meaning that services already (unfortunately) have some responsibility to implement their `stop` fns to be resilient to that... I think we have a ticket somewhere about some services being subject to an NPE in this situation today.  So I guess it doesn't feel like double-stop adds much more burden than we already have due to that issue...

camlow325: @cprice404 Fair enough.  May just need to document this as an expectation for services, then.  Arguably a "breaking change", I suppose, although it would only come into play when a new feature, HUP support, gets used.  So semantically may be okay to keep Trapperkeeper on major version 1 after this PR comes in.

So if we're willing to live with the double-stop behavior, not sure it's worth the effort to have the implementation here call `stop` on other services after the first one fails since the tk-main shutdown processing would ensure later that every service's `stop` is called at least one time.

src/puppetlabs/trapperkeeper/app.clj
```diff
@@ -12,4 +12,5 @@
                                  "occurred, return the input parameter."))
   (init [this] "Initialize the services")
   (start [this] "Start the services")
-  (stop [this] "Stop the services"))
\ No newline at end of file
+  (stop [this] "Stop the services")
+  (restart [this] "Stop and restart the services"))
```
adreyer: This should probably return the tk to for consistency with start, stop and init.

src/puppetlabs/trapperkeeper/core.clj
```diff
@@ -32,13 +32,13 @@
   (`init`, `start`, `stop`) as you see fit;  if you'd like to have the trapperkeeper
   framework block the main thread to wait for a shutdown event, call `init`,
   `start`, and then `run-app`."
-  [services config-data]
+  [services config-data-fn]
```
adreyer: This seems like a breaking change. I don't know how important it is in production uses but it breaks the user.clj in the template.

cprice404: That's fair; I meant to revisit this and possibly do some type checking on this argument.  We will either amend this PR or file a follow-up ticket that is addressed before a release, thanks for pointing it out.

test/puppetlabs/trapperkeeper/services_test.clj
```diff
@@ -1,13 +1,19 @@
...
-                [bootstrap-services-with-empty-config
-                 with-app-with-empty-config]]
+             [bootstrap-services-with-empty-config
+              with-app-with-empty-config]]
             [schema.test :as schema-test]
-            [puppetlabs.kitchensink.testutils.fixtures :refer [with-no-jvm-shutdown-hooks]]))
+            [puppetlabs.kitchensink.testutils.fixtures :refer [with-no-jvm-shutdown-hooks]]
+            [beckon]
+            [puppetlabs.trapperkeeper.internal :as internal]
+            [puppetlabs.trapperkeeper.testutils.logging :refer [with-test-logging with-test-logging-debug]]
```
nwolfe: Still have the `with-test-logging-debug` imported here

haus: thanks

src/puppetlabs/trapperkeeper/app.clj
haus: thanks. updated to do so.

haus:
I'm seeing a test failure related to the optional deps. more to come.

haus:
Ah, yes rebasing the optional-deps work in has broken those tests as they use build-app.

haus:
I'm going to see about updating build-app to take both a function and maintain previous behavior. Ideally, my last commit can go away then and existing repls will continue to work.

haus:
pushed out a commit that updates build-app to support a map or function for config-data. i've verified that the puppet-server repl works as is with these changes.

src/puppetlabs/trapperkeeper/core.clj
```diff
@@ -35,10 +35,11 @@
   [services config-data]
   {:pre  [(sequential? services)
           (every? #(satisfies? services/ServiceDefinition %) services)
-          (map? config-data)]
+          (ifn? config-data)]
    :post [(satisfies? app/TrapperkeeperApp %)]}
-  (config/initialize-logging! config-data)
-  (internal/build-app* services config-data (promise)))
+  (let [config-data-fn #(if (map? config-data) (constantly config-data) config-data)]
```
nwolfe: Do we need to add the map-or-fn support to the other similar functions below?

nwolfe: Also, do we need the anonymous function that wraps the `if map?` or can we just get rid of that like:
```clj
(let [config-data-fn (if (map? config-data) (constantly config-data) config-data] ...)
```

src/puppetlabs/trapperkeeper/core.clj
```diff
@@ -142,10 +153,12 @@
...
   {:pre [(map? cli-data)]}
   (let [app (boot-with-cli-data cli-data)]
-    ;; This line populates the `main-app` variable with the TrapperkeeperApp
-    ;; instance.  This allows it to be referenced in a remote nREPL session.
-    (alter-var-root #'main-app (fn [_] app))
-    (run-app app)))
+    ;; This adds the TrapperkeeperApp instance to the tk-apps list, so that
+    ;; it can be referenced in a remote nREPL session, etc.
+    (swap! internal/tk-apps conj app)
+    (internal/register-sighup-handler)
```
cprice404: One thing I notice here:

We don't do the initial registration of the sighup handler until after the app is booted.  I think this lessens the concerns that people had about a HUP causing indeterminate behavior if the user sent it before the app had finished booting up.  The behavior won't be indeterminate -- it will just shut down the process.

We might want to revisit that in the future, and change this up so that the registration of the handler happens earlier *and* it's smart enough not to initiate the restart while the initial startup is in progress, but I think we can defer that for now.  The fact that the behavior is not indeterminate seems to alleviate the most pressing concerns.

test/puppetlabs/trapperkeeper/services_test.clj
```diff
@@ -49,56 +55,181 @@
...
+  (testing "service contexts should be cleared out during a restart"
+    (let [test-context (atom {})
+          service1 (service EmptyService
+                     []
+                     (init [this context]
+                           (swap! test-context merge context)
+                           {:foo "bar"}))]
+      (with-app-with-empty-config app [service1]
+        (app/restart app))
+      (is (= {} @test-context)))))
```
cprice404: I think this might be easier to read if you just set the initial atom to `nil`, and used `reset!` instead of `swap!`/`merge`.

I also wouldn't mind adding a few more assertions in this test.  Like... maybe...

1. rename `test-context` to `test-init-context`
2. add a second atom that is called `test-start-context`
3. add a `start` lifecycle fn that has the same sort of `swap!` logic, to `reset!` its atom to the context that it was passed
4. Inside of the `with` block, add assertions for the value of the two atoms
5. still inside of the `with` block, reset the two atoms to `nil`
6. add an assertion on the second atom outside of the `with` block.

Bonus points for adding in one more atom that keeps a counter of the number of times `init` is called, and puts that counter into the context map that `init` returns, so that the assertions against the `start` atom can include the value of that counter.

test/puppetlabs/trapperkeeper/services_test.clj
```diff
@@ -49,56 +55,181 @@
...
+          services (conj (create-lifecycle-services call-seq) (service
+                                                                EmptyService
+                                                                []
+                                                                (stop [this context] (throw (IllegalStateException. "Exploding Service")))))]
+      (ks-testutils/with-no-jvm-shutdown-hooks
+       ; We can't use the with-app-with-empty-config macro because we don't
+       ; want to use its implicit tk-app/stop call. We're asserting that
+       ; the stop will happen because of the exception. So instead, we use
+       ; the tk/run-app here to block on the app until the restart is
+       ; called an explodes in an exception.
```
cprice404: s/an/and

test/puppetlabs/trapperkeeper/services_test.clj
```diff
@@ -49,56 +55,181 @@
...
+                           (swap! test-context merge context)
+                           {:foo "bar"}))]
+      (with-app-with-empty-config app [service1]
+        (app/restart app))
+      (is (= {} @test-context)))))
+
+(deftest test-exception-during-restart
+  (testing "restart should halt if an exception is raised"
+    (let [call-seq (atom [])
+          services (conj (create-lifecycle-services call-seq) (service
```
cprice404: nit: probably more readable to put a newline before the anonymous service definition

test/puppetlabs/trapperkeeper/services_test.clj
```diff
@@ -49,56 +55,181 @@
...
+         (is (= [:init-service1 :init-service2 :init-service3
+                 :start-service1 :start-service2 :start-service3]
+                @call-seq))
+         (app/restart app)
+         (try
+           @app-running
+           (catch ExecutionException e
+             (is (instance? IllegalStateException (.getCause e)))
+             (is (= "Exploding Service" (.. e getCause getMessage)))))
+         ))
```
cprice404: nit: bring these up to the previous line

cprice404:
I left a few minor nits, and some suggestions for making one of the tests just a bit more complete.  Other than I'm +1 and will merge this today unless anyone objects.

(There is still some additional work to do before we'd roll out a *release*, I'm just talking about merging this PR... main implication there is that we can't do any other new TK releases that don't include this, after this is merged, unless we create a new branch... which would be nice to avoid.)

@KevinCorcoran @camlow325 @pcarlisle @senior @adreyer  if you have any concerns about me merging this today once @haus has had a chance to review the last round of feedback, let me know, but I feel like I've already pretty much checked with most of you and it seems like everyone's good with it.

test/puppetlabs/trapperkeeper/services_test.clj
haus: Pushed out an updated test to do all the things

cprice404:
+1

PR #199, opened 2015-12-17
Optional Dependencies

nathanielksmith:
This PR does the following:

* Upgrades the version of schema we use
* Uses schema to validate things in services-internal
* Adds support for optional dependencies; namely, transformation of an
  alternative dependency specification into a form Prismatic Graph can
  understand and adding a helper function for determining if a service
  has been included.

This commit is purely additive; no breaking changes are introduced.

This work represents the absolute minimum of work required to make this
a usable feature; while additive improvements may be made, the things
exposed here are not expected to change.

**Important notes**

* There are some functions marked with a TODO about turning them into schemas; I think the whole validation flow in `services-internal` can be drastically simplified into a single, schemafied function with schema exception catching that implement the various IllegalArgumentExceptions. I did not have time to incorporate that refactor and can ticket it.
* In doing this work, it became apparent that it would be excellent to be able to get a reference to the TrapperKeeper reified App from a service definition. That work was a stretch goal of this ticket and sadly I ran out of time. I can also ticket that.

### DIFF
--- a/src/puppetlabs/trapperkeeper/services.clj
+++ b/src/puppetlabs/trapperkeeper/services.clj
@@ -20,8 +20,10 @@
   "Common functions available to all services"
   (service-id [this] "An identifier for the service")
   (service-context [this] "Returns the context map for this service")
-  (get-service [this service-id] "Returns the service with the given service id")
+  (get-service [this service-id] "Returns the service with the given service id. Throws if service not present")
+  (maybe-get-service [this service-id] "Returns the service with the given service id. Returns nil if service not present")
   (get-services [this] "Returns a sequence containing all of the services in the app")
+  (service-included? [this service-id] "Returns true or false whether service is included")
   (service-symbol [this] "The namespaced symbol of the service definition, or `nil`
                           if no service symbol was provided."))
 
@@ -75,12 +77,16 @@
                                             (format
                                               "Call to 'get-service' failed; service '%s' does not exist."
                                               service-id#)))))
+                             (maybe-get-service [this# service-id#]
+                               (get-in ~'@tk-app-context [:services-by-id service-id#] nil))
                              (get-services [this#]
                                (-> ~'@tk-app-context
                                    :services-by-id
                                    (dissoc :ConfigService :ShutdownService)
                                    vals))
                              (service-symbol [this#] '~service-sym)
+                             (service-included? [this# service-id#]
+                               (not (nil? (get-in ~'@tk-app-context [:services-by-id service-id#] nil))))
 
                              Lifecycle
                              ~@(si/fn-defs fns-map lifecycle-fn-names)
--- a/src/puppetlabs/trapperkeeper/services_internal.clj
+++ b/src/puppetlabs/trapperkeeper/services_internal.clj
@@ -5,12 +5,11 @@
             [schema.core :as schema]
             [puppetlabs.kitchensink.core :as ks]))
 
-(def ServiceSymbol
-  "For internal use only; this schema gives us a way to differentiate between
-  a symbol representing the name of a service, and a symbol representing the
-  name of the *protocol* for a service.  This is necessary because the `service`
-  macro accepts both as optional leading arguments when defining a service."
-  {:service-symbol (schema/pred symbol?)})
+(def optional-dep-default
+  "This value is used in place of an optional service
+   when the service is not included. In the future, it may be a more interesting or
+   useful dummy service."
+  nil)
 
 (defn protocol?
   "A predicate to determine whether or not an object is a protocol definition"
@@ -29,23 +28,67 @@
        (symbol? (first sig))
        (vector? (second sig))))
 
-(defn fns-map?
-  "A predicate to determine whether or not an object is a map of fns, as
-  used internally by the `service` macro"
-  [m]
-  (and (map? m)
-       (every? keyword? (keys m))
-       (every? seq? (vals m))
-       (every? fn-sig? (apply concat (vals m)))))
+(def Protocol
+  "A schema to determine whether or not an object is a protocol definition."
+  (schema/pred protocol?))
 
-(defn var->symbol
+(def Symbol
+  "Interal schema type for validating symbol values."
+  (schema/pred symbol?))
+
+(def Var
+  "Interal schema type for validating var values."
+  (schema/pred var?))
+
+(def ServiceFnMap
+  "Internal schema for a service's map of function references."
+  {schema/Keyword IFn})
+
+(def FnsMap
+  "Internal schema for a map of a service's function body forms."
+  {schema/Keyword [(schema/pred fn-sig?)]})
+
+(def InternalServiceMap
+  "Schema defining TK's internal representation of a service. This is used while
+  processing the service macro."
+  {:service-sym (schema/maybe Symbol)
+   :service-protocol-sym (schema/maybe Symbol)
+   :service-id schema/Keyword
+   :service-fn-map ServiceFnMap
+   :dependencies (schema/pred sequential?)
+   :fns-map FnsMap})
+
+(def ServiceMap
+  "Schema defining the map used to represent a schema in the output service
+   map."
+  {schema/Keyword IFn})
+
+(def DependencyMap
+  "Schema for a map detailing optional vs. required dependencies for a
+   service."
+  {:optional (schema/pred vector?)
+   :required (schema/pred vector?)})
+
+(schema/defn ^:always-validate var->symbol :- Symbol
   "Returns a symbol for the var, including its namespace"
-  [fn-var]
-  {:pre [(var? fn-var)]
-   :post [(symbol? %)]}
+  [fn-var :- Var]
   (symbol (str (-> fn-var meta :ns .name))
           (str (-> fn-var meta :name))))
 
+(schema/defn ^:always-validate transform-deps-map :- (schema/pred vector?)
+  "Given a map of required and optional dependencies, return a vector that
+  adheres to Prismatic Graph's binding syntax. Specifically, optional
+  dependencies are transformed into {ServiceName nil}."
+  [deps :- DependencyMap]
+  (loop [optional (:optional deps) output (:required deps)]
+    (if (empty? optional)
+      output
+      (let [dep (first optional)
+            optional-form (array-map dep optional-dep-default)]
+        (recur (rest optional) (conj output optional-form))))))
+
+;; TODO Not converting to use schema since this fn's behavior should literally
+;; be a schema
 (defn validate-fn-forms!
   "Validate that all of the fn forms in the service body appear to be
   valid fn definitions.  Throws `IllegalArgumentException` otherwise."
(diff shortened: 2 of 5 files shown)


src/puppetlabs/trapperkeeper/services.clj
```diff
@@ -22,6 +22,8 @@
   (service-context [this] "Returns the context map for this service")
   (get-service [this service-id] "Returns the service with the given service id")
   (get-services [this] "Returns a sequence containing all of the services in the app")
+  (call-service-fn [this other-service-id fn-name args default] "Calls named function in other service, returning default if that service is not included.")
+  (service-included? [this service-id] "Returns nil if service does not exist and the service if it does")
```
KevinCorcoran: I would expect that this is exactly what `get-service` already does, but since you added this, I'm assuming that `get-service` throws if the service doesn't exist?  Perhaps it should simply be changed?

KevinCorcoran: Yup, that's what `get-service` does.

In an ideal world I think it would be better for `get-service` to not throw in that case.  But, given that it's all shipped as such, this is probably reasonable.

src/puppetlabs/trapperkeeper/services.clj
```diff
@@ -22,6 +22,8 @@
   (service-context [this] "Returns the context map for this service")
   (get-service [this service-id] "Returns the service with the given service id")
   (get-services [this] "Returns a sequence containing all of the services in the app")
+  (call-service-fn [this other-service-id fn-name args default] "Calls named function in other service, returning default if that service is not included.")
```
KevinCorcoran: Perhaps it'd be nice to switch the order of `args` and `default`, which should remove the requirement that callers wrap `args` in a list ... assuming that protocols support `& args` syntax, which I am not sure if they do.

src/puppetlabs/trapperkeeper/services_internal.clj
```diff
@@ -29,23 +28,67 @@
...
-       (every? keyword? (keys m))
-       (every? seq? (vals m))
-       (every? fn-sig? (apply concat (vals m)))))
+(def Protocol
+  "A schema to determine whether or not an object is a protocol definition."
+  (schema/pred protocol?))
 
-(defn var->symbol
+(def Symbol
+  "Interal schema type for validating symbol values."
```
KevinCorcoran: typo - `Interal`

src/puppetlabs/trapperkeeper/services_internal.clj
```diff
@@ -29,23 +28,67 @@
...
-  (and (map? m)
-       (every? keyword? (keys m))
-       (every? seq? (vals m))
-       (every? fn-sig? (apply concat (vals m)))))
+(def Protocol
+  "A schema to determine whether or not an object is a protocol definition."
+  (schema/pred protocol?))
 
-(defn var->symbol
+(def Symbol
```
KevinCorcoran: Alternatively, `clojure.lang.Symbol` can just be used directly as a schema instead of defining this.

test/puppetlabs/trapperkeeper/optional_deps_test.clj
```diff
@@ -0,0 +1,187 @@
...
+                 (get-haiku (tka/get-service app :PoetryService))))))
+
+      (testing "and the named function does not exist"
+        (let [bad-poetry-service (service PoetryService
+                                          {:required []
+                                           :optional [HaikuService SonnetService]}
+                                          (get-haiku [this]
+                                                     (tks/call-service-fn this
+                                                                          HaikuService
+                                                                          :sestina
```
KevinCorcoran: This example is too trivial.  Instead of just using `:sestina` you should actually write one.  :trollface:

KevinCorcoran:
This is awesome!  The code changes are much less disruptive than I thought they might be, which is great news.

cprice404:
@nathanielksmith thanks for doing this!

I haven't had time to review yet but will do it ASAP.

One question that I have is about the final commit that adds the `call-service-fn` stuff.  w/o thorough review I'm a little iffy about that API, and have a general preference for minimizing the surface area of the API.  It seems like that addition eliminates the need for the caller to know the namespace of the protocol for the upstream service, but I'm not 100% sure that's a valuable enough addition to warrant adding the additional API?  What other benefits do you feel like that API provides?

src/puppetlabs/trapperkeeper/services.clj
nathanielksmith: Yeah, I'd rather that behavior, too. Changing that seems like we'd need an X release though, and this PR is currently a solid Y. I think I'd prefer to keep this PR a Y.

test/puppetlabs/trapperkeeper/optional_deps_test.clj
nathanielksmith: :stuck_out_tongue:

src/puppetlabs/trapperkeeper/services.clj
nathanielksmith: ah, interesting idea. I was trying to keep symmetry with (get) but this makes things much better imo

cprice404:
Also I am +1 on the creation of the two follow-up tickets that you mentioned in the description, if you haven't already.

src/puppetlabs/trapperkeeper/services.clj
nathanielksmith: As suspected, protocols don't support the & syntax, so i'm leaving this unchanged.

src/puppetlabs/trapperkeeper/services_internal.clj
nathanielksmith: I just aliased clojure.lang.Symbol to Symbol for ease of use.

nathanielksmith:
@cprice404 call-service-fn is a kind of cumbersome API; in discussing this with @KevinCorcoran and @nwolfe during hack week we all agreed it'd be nice to not have to conditionally gate calls with `(service-included?)` . However, it's definitely redundant, and if you think it detracts from the API overall then we can take it out.

nathanielksmith:
@cprice404 TK-309 and TK-310

fpringvaldsen:
\o/ This is great! I am :+1:

src/puppetlabs/trapperkeeper/services.clj
cprice404: Strongly agree with both points... that behavior would be better, but keeping this as a Y is a big enough win to suck it up.

src/puppetlabs/trapperkeeper/services_internal.clj
```diff
@@ -57,9 +100,11 @@
     {:fns fns}
     (throw (IllegalArgumentException.
              (format
-               "Invalid service definition; expected function definitions following dependency list, invalid value: '\"hi\"'"
+               "Invalid service definition; expected function definitions following dependency list, invalid value: '%s'"
```
cprice404: lolol

src/puppetlabs/trapperkeeper/services.clj
cprice404: Though... maybe it would be worth adding a `get-service-if-included` fn?

nathanielksmith: that's what service-included? does, basically...

cprice404: oh, duh.  I hadn't gotten far enough to look at the impl (or, apparently, read the docstring) yet.  how do you feel about the name of that fn?  My brain is kind of trained to assume that a fn that is named with a trailing `?` is a predicate function.

nathanielksmith: I kind of punted since the correct solution is to have get-service actually check if a service *can* be optional and only throw if it cannot. If it can, then we can write code like

`(if-let [svc (get-service this :SomeOptionalService)] ... ...)`

I'm okay changing this name if you have suggestions... get-service-if-included feels wordy, but i seem to be the only person who likes the maybe-* pattern for things that might return nil (as in `maybe-get-service`)

src/puppetlabs/trapperkeeper/services_internal.clj
```diff
@@ -346,23 +336,21 @@
...
-(defn parse-service-forms!
+(schema/defn ^:always-validate parse-service-forms! :- InternalServiceMap
   "Parse the forms provided to the `service` macro.  Return a map
   containing all of the data necessary to implement the macro:
 
   :service-protocol-sym - the service protocol symbol (or nil if there is no protocol)
   :service-id           - a unique identifier (keyword) for the service
   :service-fn-map       - a map of symbols for the names of the functions provided by the protocol
-  :dependencies         - a vector specifying the dependencies of the service
+  :dependencies         - a vector specifying the dependencies of the service. Must adhere to the fnk binding form.
```
cprice404: fnk binding form, or your new map form, right?

src/puppetlabs/trapperkeeper/services.clj
cprice404: I have no real beef with the `maybe-*` pattern.

I like the idea of being able to write `if-let` blocks, and I also like the idea of having a `service-included?` predicate, but I'd probably vote for two separate fns.  So, yeah, I'd vote for changing `service-included?` to just return a bool, and adding a `maybe-get-service` (or whatever other name).

Open to other people's input on this, though.

cprice404:
I've reviewed this now and am a big +1 on the first 3 commits, besides my nitpicky questions/comments about fn names above.

I think I'm mildly in favor of stripping off the fourth commit for now, though... the `call-service-fn` API just feels a little inconsistent to me.  I think it's partially because we would be introducing a place where you refer to fn names via keywords for the first time... and there is a part of me that kind of prefers the explicit nature of expecting callers to refer to the actual protocol for the upstream service.

It seems like it'd be better to err on the side of not including that for now, but keeping it in our back pocket... and then as we go through some real-world use-cases of implementing optional services, we'll probably have a better idea of how frequent / severe of an issue it is and we could revisit the possible API improvements at that point?  And maybe resuscitate that commit?

cprice404:
Also we will need to make sure to update the docs when we merge this.  (I wish they were still in the source tree :( )

cprice404:
Also I'm willing to discuss the `call-service-fn` thing a bit more before making a final call; if I'm the only one with hesitation about it then it certainly warrants some conversation.

src/puppetlabs/trapperkeeper/services_internal.clj
nathanielksmith: fixed

src/puppetlabs/trapperkeeper/services.clj
nathanielksmith: Updated with what we discussed.

cprice404:
I'm +1 on this now, will merge today and probably do a Y release today unless anyone objects.

cprice404:
@KevinCorcoran @fpringvaldsen you guys good?

KevinCorcoran:
:+1: 

... but if we are going to ship this in a release it would be really nice if we had docs to go along with it when the release was announced.

cprice404:
@fpringvaldsen you good?

fpringvaldsen:
@cprice404 I just looked over the changes, and I'm :+1:

cprice404:
fuckyeah

PR #200, opened 2015-12-29
(PDB-2276) Add main-with-config-hook entry point

mullr:
PuppetDB currently uses robert/hooke to glom onto `config/parse-config-data` so it can do some validation and synthesize the web routing service config in a timely manner. Aside from being generally icky, this will outright break in Clojure 1.8 if we use `-Dclojure.compiler.direct-linking=true` (which we want to, since it removes an indirection from *every* function call). 

This is one way to solve the problem, but not necessarily the only one. Please view this PR as the start of a conversation, as I'm not sure what other tk users do / need on this front. 

A less invasive approach (requiring fewer source changes, anyway) is to add `^:redef` to `config/parse-config-data`, allowing robert/hooke to keep working, but that's like going out of our way to keep things icky.

### DIFF
--- a/src/puppetlabs/trapperkeeper/core.clj
+++ b/src/puppetlabs/trapperkeeper/core.clj
@@ -93,26 +93,32 @@
   `cli-data` is a map of the command-line arguments and their values.
   `puppetlabs.kitchensink/cli!` can handle the parsing for you.
 
+  `preprocess-config-fn` is a function through which the config data is passed
+  after it is parsed. If not provided, it defaults to `identity`.
+
   Their must be a `:config` key in this map which defines the .ini file
   (or directory of files) used by the configuration service.
 
   Returns a TrapperkeeperApp instance.  Call `run-app` on it if you'd like to
   block the main thread to wait for a shutdown event."
-  [cli-data]
-  {:pre  [(map? cli-data)]
-   :post [(satisfies? app/TrapperkeeperApp %)]}
-  ;; There is a strict order of operations that need to happen here:
-  ;; 1. parse config files
-  ;; 2. initialize logging
-  ;; 3. initialize plugin system
-  ;; 4. bootstrap rest of framework
-  (let [config-data (config/parse-config-data cli-data)]
-    (config/initialize-logging! config-data)
-    (plugins/add-plugin-jars-to-classpath! (cli-data :plugins))
-    (-> cli-data
-        (bootstrap/find-bootstrap-config)
-        (bootstrap/parse-bootstrap-config!)
-        (internal/boot-services* config-data))))
+  ([cli-data preprocess-config-fn]
+   {:pre  [(map? cli-data)
+           (fn? preprocess-config-fn)]
+    :post [(satisfies? app/TrapperkeeperApp %)]}
+   ;; There is a strict order of operations that need to happen here:
+   ;; 1. parse config files
+   ;; 2. initialize logging
+   ;; 3. initialize plugin system
+   ;; 4. bootstrap rest of framework
+   (let [config-data (preprocess-config-fn (config/parse-config-data cli-data))]
+     (config/initialize-logging! config-data)
+     (plugins/add-plugin-jars-to-classpath! (cli-data :plugins))
+     (-> cli-data
+         (bootstrap/find-bootstrap-config)
+         (bootstrap/parse-bootstrap-config!)
+         (internal/boot-services* config-data))))
+  ([cli-data]
+   (boot-with-cli-data cli-data identity)))
 
 
 ;; This variable is used to hold a reference to the "main" trapperkeeper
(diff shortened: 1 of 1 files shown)


src/puppetlabs/trapperkeeper/core.clj
```diff
@@ -139,29 +145,31 @@
...
+   {:pre [(map? cli-data)
+          (fn? preprocess-config-fn)]}
+   (let [app (boot-with-cli-data cli-data preprocess-config-fn)]
+     ;; This line populates the `main-app` variable with the TrapperkeeperApp
+     ;; instance.  This allows it to be referenced in a remote nREPL session.
+     (alter-var-root #'main-app (fn [_] app))
+     (run-app app)))
+  ([cli-data] (run cli-data identity)))
+
+(defn main-with-config-hook
```
senior: Just to make sure I understand this, an example of it's usage would be PuppetDB having it's own main function:

``` clojure
(defn our-real-main [& args]
  (main-with-config-hook args pdb-specific-config-fn))
```
Then our tests would switch to the new main function and we would update our EZBake config to include this new main-namespace by setting a [config propery ](https://github.com/puppetlabs/ezbake/blob/master/src/puppetlabs/ezbake/core.clj#L405).

When I created the defaulted config service in PuppetDB I had a much uglier hack that I [pushed up here](https://github.com/senior/trapperkeeper/commit/c09bef628817257a9ae847e5d8dfdd3d058c59cb). I think this approach (if I'm understanding it correctly) is more work, but is better and it's much more obvious what's going on.

senior:
@mullr What were the speed differences you notice with the direct linking enabled?

mullr:
@senior didn't measure, just assumed it's a direction we want to go. I'm pretty sure we've all seen defn-associated var lookup rise to the top of various profiles though.

senior:
@mullr I'd be interested in seeing some numbers, I made the switch locally a month or so ago and didn't see any difference, might have been doing something wrong though

cprice404:
So... I'm not necessarily opposed to this, it definitely seems better than the robert hooke thing... but, the need for this at all seems like it highlights an inconsistency in how different teams are using the framework.  It seems like it would be better if we could just solve whatever problem was forcing you to need a hook in the first place, since this isn't a need we've come across with any of the other apps.

The last time I peeked at the puppetdb code it looked like maybe the only thing you were still using hooke for was to manage the web routes, so, personally, it seems like a better idea to me to just standardize on the way the rest of the apps are doing their web routes, and then incrementally improve that feature if it's deficient in some way...

maybe there are other things you are still doing in your hook, though?

mullr:
@cprice404 Yeah, that's totally fair. Here's what we're doing:

```clj
(def adjust-and-validate-tk-config
  (comp add-web-routing-service-config
        warn-retirements
        validate-db-settings))
```

see:
- https://github.com/puppetlabs/puppetdb/blob/master/src/puppetlabs/puppetdb/config.clj#L321
- https://github.com/puppetlabs/puppetdb/blob/master/src/puppetlabs/puppetdb/config.clj#L294
- https://github.com/puppetlabs/puppetdb/blob/master/src/puppetlabs/puppetdb/config.clj#L179

The validation stuff doesn't really need to happen here, we just did it because it's convenient. It could easily be moved to our DefaultedConfigService. 

`add-web-routing-service-config` is the reason we still need the hook. You can see from the source (the first link, above) what we need to do. It seems rather PDB-specific, honestly. Are other apps doing the same kind of thing?

cprice404:
@mullr yeah for the routing stuff, everyone else is just getting that from a normal config file.  I did some work on running PuppetDB in-process with the other PE apps last week (explanatory e-mail forthcoming), and it seemed to run fine with the routing stuff coming from the config files instead of from the hook... but I didn't do any kind of super-thorough validation.

mullr:
@cprice404 yeah, it should definitely function. The code we have written there is actually the result of a very long discussion we had about what users should be able to configure and what they should see in the config files. @ajroetker did most of that work, maybe he can speak to it.

cprice404:
Yeah, I've heard bits of conversation about what the user should and shouldn't be able to configure, and I agree with them... but since the other PE apps have this same issue it'd be cool if we could come up with a more holistic solution.  Which is why my vote would be to just put it into config files for now (like the other apps do), and then based on how terrible we think that is, work on a solution that would fix it across all of the apps.

Just my 2 cents, though.

cprice404:
Where'd we land on this?  Again, I'm not really opposed to this, just trying to sort out whether or not it's actually needed or whether there's a better way that is more consistent across all of our apps.

I'm happy to do a more thorough review if this is still needed.

mullr:
@cprice404 we decided to move puppetdb over to using a config file for this stuff, as we no longer support the user changing our base url. So we won't need this.

cprice404:
@mullr cool, thanks.

PR #203, opened 2016-01-05
(TK-311) Add log-event-listener isStarted method

rbrw:


cprice404:
yuck :)

Looks fine though, will test against puppetserver locally and merge if it goes OK

PR #206, opened 2016-01-15
(MAINT) increase timeout in service test

cprice404:
The recent changes related to HUP support have a test that
sends a HUP signal and then waits for some period of time
to validate that the restart behavior took effect.  The initial
commit of this code appears to be using too low of a timeout
for some of our CI hardware.

This commit bumps the timeout up from 1s to 5s.


nwolfe:
:+1:

PR #211, opened 2016-01-20
[discuss] Add service-apply function.

nathanielksmith:
This PR is a re-thought attempt at the `call-service-fn` I had originally added in TK-299.

I'm opening this because I came across a real usecase for this; when you want to have an optional dependency on a service without actually having to have a maven (ie, `project.clj`) dependency on that service's protocol.

This means you can be fully free from clojure namespace importation in your calling of TK functions of other services. This might be evil conceptually, but means optional dependencies can be even more of a burden-lifting feature in terms of project dependencies.

### DIFF
--- a/src/puppetlabs/trapperkeeper/services.clj
+++ b/src/puppetlabs/trapperkeeper/services.clj
@@ -35,6 +35,16 @@
 
 (def lifecycle-fn-names (map :name (vals (:sigs Lifecycle))))
 
+(defn service-apply
+  "Calls named function in other service, returning default if that service is not included."
+  [service fn-name default & args]
+  (if-let [service-fn (get service fn-name)]
+    (apply service-fn args)
+    (if (nil? service)
+      default
+      (throw (IllegalArgumentException.
+               (format "Call to 'service-apply' failed; service has no fn %s" fn-name))))))
+
 (defmacro service
   "Create a Trapperkeeper ServiceDefinition.
 
--- a/test/puppetlabs/trapperkeeper/optional_deps_test.clj
+++ b/test/puppetlabs/trapperkeeper/optional_deps_test.clj
@@ -1,6 +1,6 @@
 (ns puppetlabs.trapperkeeper.optional-deps-test
   (:require  [clojure.test :refer :all]
-             [puppetlabs.trapperkeeper.services :refer [service defservice get-service] :as tks]
+             [puppetlabs.trapperkeeper.services :refer [service defservice get-service service-apply] :as tks]
              [puppetlabs.trapperkeeper.app :as tka]
              [puppetlabs.trapperkeeper.core :refer [build-app] :as tkc]
              [schema.test :as schema-test]))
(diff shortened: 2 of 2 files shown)


cprice404:
Part of the decision to use protocols in TK was because they seemed like the main mechanism in Clojure for explicitly specifying a contract/API for an object.  That generally seems like a win to me, rather than an obstacle to be worked around.  I would be interested in hearing more about why the lein dependencies or namespace importation are so undesirable.  Without them, it sort of feels like we're just drifting back into the murky waters where things like reflection and duck-typing swim in languages like Ruby :)

Not saying I can't be convinced of this, but it sets off alarm bells for me, so I'd be interested in hearing more details about the motivation.

nathanielksmith:
@cprice404 I agree that this could lead to murky waters. My main motivation still goes back to the local development usecase when you want to work on a single service in isolation and that service is, through a miracle of packaging later on, included in some combined TK container. It seems like a win, there, to not have to actually have namespaces on disk when they are optional anyway during local development. It's less lein pedantic hell and version number wrangling when that really happens at a higher packaging level anyway.

cprice404:
If the dependency is possible (even if optional), then it seems like the repo needs to have some mechanism for doing testing against it even for local development.  Without the protocol there is no way at all to know that the functions/signatures exist until the downstream packaging occurs, which seems like way too late of a point in time to get to before having *any* test coverage of the code.

If the services are going to be composed into a container together, then the lein pedantic stuff is going to have to be dealt with sooner or later, and I'm of the opinion that sooner is better.

In a pinch, one could split the upstream protocol off into its own jar/artifact, (e.g. how slf4j has an slf4j-api artifact which only contains API definitions).  That would be a much lighter-weight dependency that could be used during development time w/o incurring quite as much of the cost of maintaining the lein pedantic stuff up front (though it obviously incurs additional cost in maintaining the new `-api` artifact).

Basically, for the local development use case that you mention, my take is that you're just moving the work of dealing with lein conflicts farther out in time, and at the same time you're potentially sacrificing some compile-time error detection and introducing risk of a certain class of runtime errors that wouldn't have been possible with the protocol-based implementation.

I'm still interested in hearing if you have other use cases in mind, and/or if anyone else has different opinions about this.

KevinCorcoran:
If we had this function, we could have used an optional dependency to solve the File Sync + CA bootstrapping problem we've been dealing with over the last few days by adding an optional dependency from the File Sync Client to the CA Service, and then calling `initialize-ssl!` (or whatev) during File Sync Client's `init`.

One thing that's sort of interesting about that particular use-case is that, in File Sync, we have pretty isolated unit tests which tests that the SSL stuff is initialized correct, and other than that it does not get exercised in tests until it is integrated into Puppet Server.

Another thing that's perhaps worth considering is that, if downstream projects want to use this pattern, even if we choose to exclude `service-apply`, there's nothing preventing them from putting that function in a downstream codebase and using it anyway.

All that said, I don't have a strong opinion on this.  I can see both arguments.  At the end of the day, I tend to think that this would make optional dependencies more useful and it's probably not quite as evil as the current discussion seems to imply.  So I wouldn't mind including it.

But, to recycle one of my favorite quotes: the first rule of API design is: **when in doubt, leave it out**.  So (now is the part of this comment wherein I contract myself) I might suggest we sit on this for a bit and see what other use cases come up.

cprice404:
@KevinCorcoran isn't there a path forward on the file sync thing where you just refer to the CA protocol directly, via a dev dependency or something?  And then if you need an alternate implementation of that protocol for testing, you could just have a defservice in the test code?

I think we're likely to end up pulling the CA stuff out into its own repo at some point anyway, and if we did that (or do it more proactively), we could set up the artifacts in such a way as to make that protocol accessible without dragging in other dependencies that we don't want?  I guess to me this just seems more like a code smell of how we have the current services organized, rather than a shortcoming in the TK implementation.

KevinCorcoran:
Fair enough.  I think it would be possible to use a dev dependency as-is.  Not 100% sure that would work, but I think it would.  Even if it did work, it seems funky (although not totally unacceptable) for File Sync to have a dependency on Puppet Server ... much more reasonable if it's just the CA as a separate artifact.

cprice404:
Totally agree about the funk of the dependency from FS to PS.  I just think that's our fault and not TK's.  :)

cprice404:
AFAICT there is enough uncertainty about the necessity / desirability of this to where the prudent thing seems to be to fall back to the Corcoran Principle: "when in doubt, leave it out".  We can always resuscitate this branch if/when a concrete use case makes it necessary.

Any objections to me closing this for now?  @nathanielksmith @KevinCorcoran ?

KevinCorcoran:
no objections

PR #212, opened 2016-01-30
(TK-320) Add log4j compatibility layer

pcarlisle:
Add log4j-over-slf4j dependency, which provides a log4j interface for slf4j so
libraries written to that interface log correctly with logback.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -15,6 +15,7 @@
                  [org.clojure/tools.nrepl "0.2.3"]
                  [org.clojure/tools.macro "0.1.2"]
                  [ch.qos.logback/logback-classic "1.1.1"]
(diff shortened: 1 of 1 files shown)


project.clj
```diff
@@ -15,6 +15,7 @@
                  [org.clojure/tools.nrepl "0.2.3"]
                  [org.clojure/tools.macro "0.1.2"]
                  [ch.qos.logback/logback-classic "1.1.1"]
+                 [org.slf4j/log4j-over-slf4j "1.7.6"]
```
pcarlisle: This is several versions out of date, but it has no conflicts with our logback deps. Could have gone with 1.7.14 and an exclusion. Logback is also slightly old but bumping that wouldn't bring the deps that far forward.

PR #213, opened 2016-02-05
(TK-300) use clojure agent for lifecycle fns

cprice404:
This commit changes the handling of the TK lifecycle functions so that they are run on a Clojure agent.  This change is in place for both the "normal" (main, startup) code path of the life cycle functions as well as the HUP restarting.  This ensures that lifecycle functions will not be executed on the JVM signal handler thread, and that there will never be more than one thread attempting to execute lifecycle fns at a time.

### DIFF
--- a/src/puppetlabs/trapperkeeper/app.clj
+++ b/src/puppetlabs/trapperkeeper/app.clj
@@ -1,4 +1,18 @@
-(ns puppetlabs.trapperkeeper.app)
+(ns puppetlabs.trapperkeeper.app
+  (:require [schema.core :as schema]
+            [puppetlabs.trapperkeeper.services :as s]))
+
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+;;; Schema
+
+(def TrapperkeeperAppContext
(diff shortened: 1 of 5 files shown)


cprice404:
There are a bunch of cleanup commits before the real work begins... the significant changes are all in the final two commits.  May be easiest to review one commit at a time.

@camlow325 @haus no huge rush but you two gents are probably the best reviewers for this.

KevinCorcoran:
Should there be any new tests written to cover the last two commits?

cprice404:
For the commit that introduces the agent to the main thread, I do not think any additional tests are necessary.  There are tons of tests (especially in the shutdown code) that validate that the threading logic for that wasn't changed from an end-user perspective.

For the final commit, where we switch HUP to use the agents, a test *might* be possible, but so far I haven't thought of a way to write one that wouldn't be some combination of "insanely complicated", racy, or so pathological as to not really be valuable.  I'll think about it a bit more, though.

cprice404:
Attempting to use a SNAPSHOT build of this with Puppet Server and not sure it's working properly.  Stay tuned.

KevinCorcoran:
I hear ya - hard stuff to test.  But, just looking at those commits ... it's not immediately clear from the code alone what purpose they serve.  Maybe just some additional docstrings and/or comments about the "why?" would help.

cprice404:
Closing this because it has a fatal flaw, will open another PR soon.

PR #214, opened 2016-02-05
(TK-300) use core.async to run lifecycle fns

cprice404:
There are some limitations to clojure agents that made them unsuitable for this task, so here's a new PR that uses core.async instead.

The first several commits are just cleanup/schema/whitespace stuff.  The interesting stuff is in the final commit.  Probably easiest to review one commit at a time.


cprice404:
OK, I've tested this with puppet server now.  It works, except for a few tests where we were using app-context in a very evil way that is slightly broken in the new TK... but I have a patch that fixes those tests, and it works fine with this TK.  So I think this is ready for review.  I'll submit a PR for the puppetserver bits before I remove the 'FOR REVIEW' comments, though.

src/puppetlabs/trapperkeeper/internal.clj
```diff
@@ -265,18 +295,13 @@
...
+     ; that this function ; never throw anything (like a schema validation error),
      ; since it is likely to just get lost in a `future`.  Instead,
      ; invalid arguments will simply cause the shutdown promise to be delivered.
-     (assert (instance? Atom app-context))
-     (assert (map? @app-context))
-     (assert (keyword? svc-id))
-     (assert (contains? @app-context svc-id))
-     (assert (ifn? f))
-     (assert ((some-fn nil? ifn?) on-error-fn))
+     (schema/validate a/TrapperkeeperAppContext @app-context)
```
camlow325: Seems like we would be losing some of the validation that was being done previously - i.e., that the `svc-id`, `f`, and `on-error-fn` arguments are all of the type we want them to be.  Is that a problem?

src/puppetlabs/trapperkeeper/internal.clj
```diff
@@ -138,7 +141,7 @@
         required    []]
     (first (cli! cli-args specs required))))
 
-(defn run-lifecycle-fn!
+(schema/defn ^:always-validate run-lifecycle-fn! :- a/TrapperkeeperAppContext
```
camlow325: I wasn't sure if we have anything calling this which expects it to return anything - as opposed to the input `app-context` just being updated as a side effect of calling this function.  Are you expecting that the return value from this function is something that a caller should formally expect to receive back?  If not, would it be better to just not document this - so that the implementation is theoretically "freer" to be changed to do something else later if we want to?

project.clj
```diff
@@ -1,6 +1,6 @@
 (def ks-version "1.0.0")
 
-(defproject puppetlabs/trapperkeeper "1.2.1-SNAPSHOT"
+(defproject puppetlabs/trapperkeeper "1.3.0-SNAPSHOT"
```
camlow325: Are you thinking this will be a Y version bump b/c of the introduction of the core.async dependency?

src/puppetlabs/trapperkeeper/internal.clj
```diff
@@ -181,9 +183,15 @@
...
                       depends on will have their corresponding lifecycle fn
                       called first.)"
-  [app-context lifecycle-fn lifecycle-fn-name ordered-services]
+  [app-context :- (schema/atom a/TrapperkeeperAppContext)
+   lifecycle-fn :- IFn
+   lifecycle-fn-name :- schema/Str
+   ordered-services :- a/TrapperkeeperAppOrderedServices]
   (try
     (doseq [[service-id s] ordered-services]
+      (log/debugf "Running lifecycle function '%s' for service '%s'"
```
camlow325: Thanks for doing this -- seems handy.

src/puppetlabs/trapperkeeper/internal.clj
```diff
@@ -193,15 +201,37 @@
...
   "Call restart on all tk apps."
   [apps]
   (log/debug "SIGHUP handler restarting TK apps.")
-  (locking sighup-lock
-    (doseq [app apps]
-      (a/restart app))))
+  (doseq [app apps]
+    (let [{:keys [lifecycle-channel]} @(a/app-context app)
+          restart-fn #(a/restart app)]
+      (async/>!! lifecycle-channel restart-fn))))
```
camlow325: Should all of the apps that need to be restarted be restarted within one "task" handled on the lifecycle channel - as opposed to one "task" per app as it would appear to be done here.  Not sure if there would be issues at some point with other tasks potentially being interleaved between individual app restarts that could cause a problem - even if not now, then maybe for some future lifecycle task that we might introduce?

camlow325:
I would still be interested in seeing at least one test that somehow validates that the `boot-services**` and `restart` calls have to be performed serially on the async channel thread.  That said, I can't think of a way to do that which wouldn't lead to a fairly scary-looking test with timing-related dependencies that may lead to frequent false failures, so maybe not worth it?

camlow325:
:+1: Did a little functional testing of this w/ it integrated into Puppet Server and it seemed to do what I expected.  I can still do HUPs and have configuration reloaded the way I expect and different forms of triggering shutdowns seem to result in the process getting shutdown completely / not hanging.

I left a few comments but nothing too critical.

Probably want to have @haus take a second look at all of this as well?

src/puppetlabs/trapperkeeper/internal.clj
cprice404: good call, I may as well add back in some schema/validate lines for those.

src/puppetlabs/trapperkeeper/internal.clj
cprice404: Yeah, that's fair, I just annotated it because that's what it was returning, but I think I agree with your take.  Will remove.

project.clj
cprice404: Nah, because of the HUP support... doesn't that seem like a new feature that should qualify as a Y?

src/puppetlabs/trapperkeeper/internal.clj
cprice404: In this implementation each app has its own lifecycle channel, and they could actually be restarted in parallel.  I don't feel too strongly about it but I think I'd rather keep it that way until we have a reason not to; it's not really visible in the API so I think we could change it later if we needed.

project.clj
camlow325: Yeah, that makes sense.  Sorry, was thinking we'd already done a T/K release with the earlier form of HUP.  Losing track of things....

cprice404:
Added a commit that addresses the schema questions @camlow325 raised.  I'll spend a bit more time thinking about how I could write a test but I'm not that optimistic atm.

camlow325:
:+1:

cprice404:
@camlow325 I added some more commits.  a couple more simple cleanup commits, and then one to refactor boot-services* to make it more amenable to testing, and then one that adds a test for restart during boot-up.  The test ended up being not entirely horrible... I think.

haus:
I'm :+1: here. Nothing looks scary. I don't understand the macros, but everything else made sense.

camlow325:
:+1: with just the nit on unused namespaces in the new `test-queued-restarts` test.  Thanks for writing that test, btw - looks good.

cprice404:
Added a commit to remove the unused namespaces.

camlow325:
:+1:

PR #218, opened 2016-02-10
(TK-319) Handle optional dependencies w/o protocol

nathanielksmith:
Fixes a bug where the `service` macro would fail if not using a service
protocol but using the new optional dependencies binding form.

### DIFF
--- a/src/puppetlabs/trapperkeeper/services_internal.clj
+++ b/src/puppetlabs/trapperkeeper/services_internal.clj
@@ -139,6 +139,7 @@
   (let [f (first forms)]
     (cond
       (symbol? f) (merge {:service-protocol-sym f} (validate-deps-form! (rest forms)))
+      (map? f) (merge {:service-protocol-sym nil} (validate-deps-form! forms))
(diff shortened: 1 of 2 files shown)


fpringvaldsen:
:+1:

src/puppetlabs/trapperkeeper/services_internal.clj
```diff
@@ -139,6 +139,7 @@
   (let [f (first forms)]
     (cond
       (symbol? f) (merge {:service-protocol-sym f} (validate-deps-form! (rest forms)))
+      (map? f) (merge {:service-protocol-sym nil} (validate-deps-form! forms))
```
cprice404: Just to verify, there is currently no other legal syntax where a map would end up in this position, except for the optional dependencies case.  Correct?

nathanielksmith: correct. do you want further validation?

cprice404: Nah, I think that when we port this fn to schema as you described in your comment, we'll be able to take care of that then.  So I'm good with this for now.

PR #221, opened 2016-03-14
(MAINT) add testutils bootstrap-services-with-config

cprice404:
All of the various `with-app*` macros in the tk bootstrap testutils
have a corresponding `bootstrap-services*` fn that can be used in
cases where a test needs more control over the lifecycle of the app,
except for `with-app-with-config`.  This commit simply adds a
`bootstrap-services-with-config` to achieve parity there.


fpringvaldsen:
:+1:

PR #222, opened 2016-03-14
(FOR REVIEW) (TK-326) create max number of queued hup events

cprice404:
FOR REVIEW: this builds on PR #221, only the final commit of this PR is interesting.  Will rebase after that is merged.

-------------------------

This commit changes the HUP/restart logic so that if there are too
many events queued up in the core.async lifecycle channel, we won't
keep piling more on.  We instead just log a warning and ignore
the HUP request.

### DIFF
--- a/src/puppetlabs/trapperkeeper/internal.clj
+++ b/src/puppetlabs/trapperkeeper/internal.clj
@@ -18,6 +18,8 @@
 ;; (such as signal handling).
 (def tk-apps (atom []))
 
+(def max-pending-lifecycle-events 5)
+
 (defn service-graph?
   "Predicate that tests whether or not the argument is a valid trapperkeeper
   service graph."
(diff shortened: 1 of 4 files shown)


fpringvaldsen:
:+1:

src/puppetlabs/trapperkeeper/internal.clj
```diff
@@ -231,7 +233,9 @@
   (doseq [app apps]
     (let [{:keys [lifecycle-channel]} @(a/app-context app)
           restart-fn #(a/restart app)]
-      (async/>!! lifecycle-channel restart-fn))))
+      (when-not (async/offer! lifecycle-channel restart-fn)
+        (log/warnf "Too many SIGHUP restart requests queued (%s); ignoring!"
```
rlinehan: If someone has log level at `warn` or `info` then they won't see the message from https://github.com/puppetlabs/trapperkeeper/pull/222/commits/9c017b830ffe86415c6d329d19262d112e4f01b4#diff-452afe9e02bf063094d280de34cf34bcR232 that says it's gotten a SIGHUP. If that is the case, then there isn't really an antecedent to `ignoring!` saying what it is ignoring. Might be better to say `ignoring new SIGHUP restart request` or something.

rlinehan:
Left one small nitpick, but :+1: and happy to merge after this has been rebased (which I think it can be since #221 has been merged).

camlow325:
:+1:

camlow325:
Appears to have been rebased so merging...

src/puppetlabs/trapperkeeper/internal.clj
cprice404: I agree with this, will try to remember to fix in a subsequent PR.

PR #223, opened 2016-03-14
(TK-334) debug logging for restart and shutdown

cprice404:
This commit adds/improves some debugging statements to clarify what
is happening with lifecycle events inside of TK, and when we shut
down the Clojure agent pool.
    
To achieve this we needed to provide a bit of schema around the
items that are put onto the core.async channel for lifecycle management.

### DIFF
--- a/src/puppetlabs/trapperkeeper/internal.clj
+++ b/src/puppetlabs/trapperkeeper/internal.clj
@@ -255,7 +255,7 @@
       (when-not (async/offer! lifecycle-channel
                               {:type :restart
                                :task-function restart-fn})
-        (log/warnf "Too many SIGHUP restart requests queued (%s); ignoring!"
+        (log/warnf "Ignoring new SIGHUP restart requests; too many requests queued (%s)"
                    max-pending-lifecycle-events)))))
 
 (defn register-sighup-handler
--- a/src/puppetlabs/trapperkeeper/internal.clj
+++ b/src/puppetlabs/trapperkeeper/internal.clj
@@ -12,6 +12,18 @@
             [clojure.core.async :as async]
             [clojure.core.async.impl.protocols :as async-prot]))
 
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+;; Schemas
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+
+(def LifeCycleTask
+  {:type (schema/enum :restart :shutdown :boot)
+   :task-function (schema/maybe IFn)})
+
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+;; Private
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+
 ;; This is (eww) a global variable that holds a reference to all of the running
 ;; Trapperkeeper apps in the process. It can be used when connecting via nrepl
 ;; to allow you to do useful things, and also may be used for other things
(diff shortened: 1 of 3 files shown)


fpringvaldsen:
:+1:

src/puppetlabs/trapperkeeper/internal.clj
```diff
@@ -212,15 +224,22 @@
...
+                        :done)
+
+                      #{:boot :restart}
+                      (do
+                        (log/debugf "Lifecycle worker executing %s lifecycle task." type)
+                        (task-function)
+                        (log/debugf "Lifecycle worker completed %s lifecycle task; awaiting next task." type)
+                        (recur))
+
+                      (throw (IllegalStateException. "Unrecognized lifecycle task:" task))))))
```
rlinehan: It doesn't seem like it will be possible to get here, since the schema/validate on line 227 (https://github.com/puppetlabs/trapperkeeper/pull/223/commits/59347aa16f46a61fa77a41d4518e9fe728543f29#diff-452afe9e02bf063094d280de34cf34bcR227) will validate that `:type` is one of `:shutdown`, `:boot`, and `:restart`.

src/puppetlabs/trapperkeeper/internal.clj
```diff
@@ -210,15 +224,22 @@
...
+                    (condp #(contains? %1 %2) type
+                      #{:shutdown}
+                      (do
+                        (log/debug "Received shutdown command, lifecycle worker exiting.")
+                        :done)
+
+                      #{:boot :restart}
+                      (do
+                        (log/debugf "Lifecycle worker executing %s lifecycle task." type)
+                        (task-function)
```
camlow325: It looks like the schema would allow for the `task-function` to be optional.  If that's intentional, should we check that `task-function` has a non nil value before executing it?  Assuming `task-function` should be required for a `:boot` or `:restart` task, should we harden the schema to ensure that it has a non nil value?

camlow325:
:+1: although it looks like we'll probably want to rebase this on top of the previously merged #222 since some of the changed code between this one and that one look pretty similar.

src/puppetlabs/trapperkeeper/internal.clj
cprice404: I agree, but I still consider it best practice to have a default case in a switch statement.  Seems harmless?

Maybe someone changes the schema in the future and doesn't update this switch, and the default case becomes reachable?

src/puppetlabs/trapperkeeper/internal.clj
cprice404: Good catch.  In this PR, the 'shutdown' path doesn't have a task-function, which is why I had to do the schema that way.  In a later PR when I move the shutdown work into the go loop, the shutdown case gains a task-function.  So I think what would make sense to me is to leave this as-is in this PR but then just help make sure that I harden the schema in the follow-up PR?

cprice404:
Rebased and added a commit addressing @rlinehan 's comment from my previous PR.  I'm not sure I think there's anything immediately actionable from the other comments on this PR, but happy to revisit if anyone disagrees with my thoughts above.

src/puppetlabs/trapperkeeper/internal.clj
rlinehan: Yeah, that makes sense. I guess I hope someone would catch this in a review, but there's not a guarantee of that, and it's probably better to guard against it since it isn't hard.

rlinehan:
:+1: but looks like travis failed with some legitimate failures because of the log statement changing

```
FAIL in (test-max-queued-restarts) (internal_test.clj:89)

Missing expected log message when too many HUP requests queued

expected: (logged? (format "Too many SIGHUP restart requests queued (%d); ignoring!" internal/max-pending-lifecycle-events) :warn)

  actual: (logged [#<LoggingEvent [DEBUG] SIGHUP handler restarting TK apps.> #<LoggingEvent [WARN] Ignoring new SIGHUP restart requests; too many requests queued (5)>])
```

cprice404:
dang it

cprice404:
squashed in a fix for the test relating to the changed log message.

src/puppetlabs/trapperkeeper/internal.clj
camlow325: Fair enough, will try to remember to look for that.

camlow325:
Looks like Travis is passing and I don't see any other outstanding comments so merging this...

PR #224, opened 2016-03-15
(TK-330) move main shutdown logic to core async worker

cprice404:
This commit moves the "main" shutdown/stop logic off of the main
thread and onto the core.async worker thread.  This ensures that
we don't end up in a situation where we have restarts queued up
and then an error occurs and causes the situation where we have
lifecycle activity going on on the async worker thread while
the main thread is also trying to perform the shutdown tasks.

### DIFF
--- a/test/puppetlabs/trapperkeeper/shutdown_test.clj
+++ b/test/puppetlabs/trapperkeeper/shutdown_test.clj
@@ -388,6 +388,8 @@
                       context)
                (stop [this context]
                      (swap! lifecycle-events conj :stop)
+                     ;; request shutdown, which will trigger the shutdown logic
+                     ;; on the main thread.
                      (request-shutdown)
                      (deliver first-stop-begun? true)
                      @stop-should-proceed?
@@ -397,7 +399,11 @@
                {})
           ;; this ensures that the 'main' shutdown logic will be runnable,
           ;; and gives us a way to observe when it has completed.
-          app-main-thread (future (tk/run-app app))]
+          app-main-thread (future (tk/run-app app))
+          shutdown-service (tk-app/get-service app :ShutdownService)]
+
+      ;; no shutdown requested yet
+      (is (nil? (internal/get-shutdown-reason shutdown-service)))
 
       ;; now we trigger a restart, which will call 'stop' for the first time,
       ;; which will request a shutdown but will block on the stop-should-proceed
@@ -407,17 +413,17 @@
       ;; wait until we know that the shutdown has been requested
       @first-stop-begun?
 
-      ;; at this point, the app's shutdown-reason-promise should be set, but
-      ;; the main thread should be blocked because our first 'stop' is blocked.
-      (is (realized? (:shutdown-reason-promise @(tk-app/app-context app))))
-      (is (not (realized? app-main-thread)))
+      ;; we want to give the main thread a little time in case it is going to
+      ;; do anything nefarious (as it would have before we consolidated the
+      ;; shutdown logic into the core.async worker loop)
+      (Thread/sleep 100)
+
+      ;; at this point, the app's shutdown-reason-promise should be set
+      (is (not (nil? (internal/get-shutdown-reason shutdown-service))))
 
       ;; validate that the first three events are as expected (we're still blocked
       ;; in the first 'stop').
-      (let [expected-lifecycle-events [:init :start :stop]]
-        (while (< (count @lifecycle-events) (count expected-lifecycle-events))
-          (Thread/yield))
-        (is (= expected-lifecycle-events @lifecycle-events)))
+      (is (= [:init :start :stop] @lifecycle-events))
 
       ;; main thread should still be blocked
       (is (not (realized? app-main-thread)))
--- a/test/puppetlabs/trapperkeeper/shutdown_test.clj
+++ b/test/puppetlabs/trapperkeeper/shutdown_test.clj
@@ -9,7 +9,9 @@
             [puppetlabs.trapperkeeper.testutils.bootstrap :refer [bootstrap-services-with-empty-config]]
             [puppetlabs.trapperkeeper.testutils.logging :as logging]
             [puppetlabs.kitchensink.testutils.fixtures :refer [with-no-jvm-shutdown-hooks]]
-            [schema.test :as schema-test]))
+            [schema.test :as schema-test]
+            [puppetlabs.trapperkeeper.testutils.bootstrap :as testutils]
+            [puppetlabs.trapperkeeper.app :as tk-app]))
 
 (use-fixtures :once with-no-jvm-shutdown-hooks schema-test/validate-schemas)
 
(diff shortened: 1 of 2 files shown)


test/puppetlabs/trapperkeeper/shutdown_test.clj
```diff
@@ -370,3 +372,64 @@
...
+               [[:ShutdownService request-shutdown]]
+               (init [this context]
+                     (swap! lifecycle-events conj :init)
+                     context)
+               (start [this context]
+                      (swap! lifecycle-events conj :start)
+                      context)
+               (stop [this context]
+                     (swap! lifecycle-events conj :stop)
+                     (request-shutdown)
```
fpringvaldsen: Is this call what's causing the second shutdown?

fpringvaldsen:
I just had one comment, other than that, :+1:

camlow325:
Do we know what will happen if `shutdown!` is called from the shutdown hook that we register with Java - like via hitting CTRL-C on a foreground process or doing a SIGTERM/SIGKILL on a running process?  I had been under the impression in that state that the JVM might have killed off all other running threads besides the one processing the shutdown hook - in which case maybe the core.async thread wouldn't be alive anymore to handle the shutdown message?

camlow325:
At the end of this string of PRs as it currently stands, I wonder if we're going to still have a potential race condition around double-stops.

As an example, against the current code on this PR, I ran the following code in a REPL, using the `initialize-lifecycle-worker!`, stubbing in the minimal core.async calls that I think could happen if `shutdown!` were called twice:

```clj
(let [channel (async/chan)
      a-promise (promise)
      worker (initialize-lifecycle-worker
              channel
              a-promise)]
  (println "send to channel one")
  (async/>!! channel {:type :shutdown
                      :task-function nil})
  (println "wait for worker one")
  (async/<!! worker)
  (println "wait for promise one")
  @a-promise
  (println "send to channel two")
  (async/>!! channel {:type :shutdown
                      :task-function nil})
  (println "wait for worker two")
  (async/<!! worker)
  (println "wait for promise two")
  @a-promise
  (println "done"))
```

The following output appeared:

```
send to channel one
wait for worker one
wait for promise one
send to channel two
```
The REPL hung at the second call to `async/<!! channel {:type :shutdown :task function nil}`.

Is it still possible that shutdown could be called twice - like if a tk/restart and a shutdown-on-error / request-shutdown were to occur around the same time.  If that did happen, would this mean that the second shutdown could hang on its attempt to send something to the core.async channel if the channel is shutdown just prior to that send?

camlow325:
For the code sample I listed above, I am able to avoid the hang if I insert this line into the `#{:shutdown}` block inside of `initialize-lifecycle-worker!`:

```clj
(async/close! lifecycle-chan)
```
This might reduce the likelihood of a lock up but I don't know if doing the `async/>!!` would return immediately if the `close!` didn't happen before the second `async/>!!` call was made.  We may want to put logic in to close the input channel anyway after a `:shutdown` task has been pulled out of the queue, though.

test/puppetlabs/trapperkeeper/shutdown_test.clj
```diff
@@ -370,3 +372,64 @@
...
+      ;; which will request a shutdown but will block on the stop-should-proceed
+      ;; promise
+      (internal/restart-tk-apps [app])
+
+      ;; wait until we know that the shutdown has been requested
+      @first-stop-begun?
+
+      ;; at this point, the app's shutdown-reason-promise should be set, but
+      ;; the main thread should be blocked because our first 'stop' is blocked.
+      (is (realized? (:shutdown-reason-promise @(tk-app/app-context app))))
```
camlow325: Another way to do this would be to call the `get-shutdown-reason` function from the `ShutdownService` and confirm that the result is not-nil.  This seems fine as well, though.

test/puppetlabs/trapperkeeper/shutdown_test.clj
```diff
@@ -370,3 +372,64 @@
...
+      (is (not (realized? app-main-thread)))
+
+      ;; validate that the first three events are as expected (we're still blocked
+      ;; in the first 'stop').
+      (let [expected-lifecycle-events [:init :start :stop]]
+        (while (< (count @lifecycle-events) (count expected-lifecycle-events))
```
camlow325: Do we need the loop here?  Shouldn't `expected-lifecycle-events` have exactly the `:init`, `:start`, and `:stop` elements at the point the `first-stop-begun` promise is first delivered?  Could just do the lifecycle event assertion and the `not (realized? app-main-thread)` assertion one time instead of two here?

camlow325: I'm not generally a fan of sleeps but I wonder if one would be useful here to provoke the likelihood of the contention for double-stops that we're trying to provoke here?  For example, if it were possible for two stops from different threads to be running at the same time -- current code prior to the work in this PR -- a short sleep (50 milliseconds maybe?) might provoke the likelihood that we've progressed past the shutdown promise being set and into the actual `stop` call on the service being executed from the TK main thread.

camlow325:
Done reviewing for now.  I'm generally okay with the direction of the work but had a few questions around potential race conditions around calls to `shutdown!` that might be initiated from different threads concurrently and how we'll be able to coordinate that with the underlying core.async worker thread.

test/puppetlabs/trapperkeeper/shutdown_test.clj
cprice404: Correct.

cprice404:
> Do we know what will happen if shutdown! is called from the shutdown hook that we register with Java - like via hitting CTRL-C on a foreground process or doing a SIGTERM/SIGKILL on a running process? I had been under the impression in that state that the JVM might have killed off all other running threads besides the one processing the shutdown hook - in which case maybe the core.async thread wouldn't be alive anymore to handle the shutdown message?

Is that actually a thing?  I feel like we'd probably have other problems if that's how the JVM behaved.  :sheep: 

> At the end of this string of PRs as it currently stands, I wonder if we're going to still have a potential race condition around double-stops.

This chain of PRs isn't the end, it's just all that I've had time to do so far.  I'm at least still missing TK-331, and I think that might change some of what you're asking about?

In that ticket I intend to add a second channel for shutdown messages so that they can be prioritized above restart messages.  I think the concerns you've raised are valid but I think they can be resolved by a) not using a buffered channel for that channel, meaning that there can only ever be one work item queued, and b) changing the `shutdown!` logic to do a non-blocking put, and bailing if it fails to put the item onto the queue?

> (async/close! lifecycle-chan)

Yeah, that's a good call, that should be in there.  I can add that in this PR.

test/puppetlabs/trapperkeeper/shutdown_test.clj
cprice404: That seems better.

test/puppetlabs/trapperkeeper/shutdown_test.clj
cprice404: Yeah I can probably get rid of that loop.

As for the sleep; this test failed very consistently for me without the corresponding patches to the production code, even with the yield and not a sleep.  But I don't mind adding one.  To be clear though, you're saying to do that outside of any loop?

cprice404:
Thanks for the review, definitely a few actionable things here, but will hold off on making these changes until #223 gets merged.

camlow325:
RE: the question I had about the shutdown hook...

I don't have any hard evidence that when a shutdown hook fires that other threads would be dormant.  That's mostly a relic in my head from early discussions I remember having about how that all was working - and I could be totally off-base on that.  I do think it would be good to at least do some anecdotal testing of this to see what happens in that case - i.e., that the process still shuts down in a clean way from a CTRL-C or SIGTERM/SIGKILL.  If it does, I'm good with this.

---

RE: the questions I had about the potential for double-stops that might exist after this PR lands, I think we can prove that if the `async/>!!` call that is made to put a "task" onto the upstream channel will not block indefinitely - i.e., in the case that the channel is closed in the interim before/after the corresponding data is accepted into the queue, then I'm satisified with where this stands.  I'd be fine with the idea of, if necessary, having some of the work to make that happen not be done until [TK-331](https://tickets.puppetlabs.com/browse/TK-331) is completed.

test/puppetlabs/trapperkeeper/shutdown_test.clj
cprice404: @camlow325 ^^

cprice404:
This is how I'm currently reading the comments:

Stuff to do on this PR:
* Use `get-shutdown-reason` in that test
* Get rid of unnecessary loop in test
* Possibly add sleep to test

Postpone until the TK-331 PR is ready:
* add the `async/close` call inside the shutdown handler
* ensure that if two threads try to put a shutdown request into the work queue that only one will be handled, and that the other will not block indefinitely
* manually test CTRL-C

@camlow325 that seem accurate?

camlow325:
@cprice404 Yep, that all sounds right.  Thanks for the summary.

cprice404:
rebased and added a commit that addresses the issues I was aware of.

camlow325:
:+1:

cprice404:
@fpringvaldsen this look ok to you at this point?

fpringvaldsen:
:+1:

camlow325:
:ship:

PR #225, opened 2016-03-16
(TK-342) Add with-logged-event-maps

rbrw:
Add with-logged-event-maps and with-logger-event-maps for the likely
case where you want to look at event maps rather than raw log events,
and you're happy to have the wrapper handle setting up the destination
atom:

  (with-logged-event-maps events
    (log/error "foo")
    (is (some #(= "foo" (:message %)) @events)))

Thanks to Chris Price for the suggestion.


cprice404:
@rbrw this looks great to me, thanks a ton for submitting.

Would you mind adding a quick blurb to the docs to show the example usage?:
https://github.com/puppetlabs/trapperkeeper/blob/master/documentation/Test-Utils.md#logging

rbrw:
@cprice404 will do.

cprice404:
+1, merging today unless anyone else chimes in

PR #226, opened 2016-03-17
(SERVER-893) Adopt pl-clojure-style

justinstoller:
This commit adds cljfmt as a valid lein alias which will defer to the
plugin lein-cljfmt using the pl-clojure-style repo's styles.

We use a git submodule approach to pulling in pl-clojure-style to ease
the ability in the future of using the tool within an automated process
(e.g. travis-ci) and because of significant use of submodules in other
puppet labs clj projects.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -48,7 +48,7 @@
   ;; code that we have.
   :classifiers [["test" :testutils]]
 
-  :profiles {:cljfmt {:plugins [[lein-cljfmt "0.3.0"]
+  :profiles {:cljfmt {:plugins [[lein-cljfmt "0.5.0"]
                                 [lein-parent "0.2.1"]]
                       :parent-project {:path "ext/pl-clojure-style/project.clj"
                                        :inherit [:cljfmt]}}
--- a/project.clj
+++ b/project.clj
@@ -39,7 +39,8 @@
                                      :sign-releases false}]]
 
   ;; Convenience for manually testing application shutdown support - run `lein test-external-shutdown`
-  :aliases {"test-external-shutdown" ["trampoline" "run" "-m" "examples.shutdown-app.test-external-shutdown"]}
+  :aliases {"cljfmt" ["with-profile" "+cljfmt" "cljfmt"]
+            "test-external-shutdown" ["trampoline" "run" "-m" "examples.shutdown-app.test-external-shutdown"]}
 
   ;; By declaring a classifier here and a corresponding profile below we'll get an additional jar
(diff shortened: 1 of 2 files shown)


project.clj
```diff
@@ -40,7 +41,11 @@
   ;; code that we have.
   :classifiers [["test" :testutils]]
 
-  :profiles {:dev {:source-paths ["examples/shutdown_app/src"
+  :profiles {:cljfmt {:plugins [[lein-cljfmt "0.3.0"]
```
justinstoller: There's newer versions of this plugin, but every project that I found in PL that has implemented this feature has used this version (its the version used in the example).

justinstoller:
I've opened a PR with the application of `lein cljfmt fix` in #227. Separated out so we can discuss the approach (submodules) to integrating the tool here and the effects of the tool there.

fpringvaldsen:
:see_no_evil:

project.clj
pcarlisle: i believe 0.3.0 was the newest when any of those project integrated this, 0.4.0 is less than a month old

pcarlisle:
this seems like a good approach

cprice404:
@justinstoller The current flurry of TK PRs is finished.  @jpinsonault may start working on something new tomorrow or the day after, but I think that if you got these updated during the evening PDX time tonight, they could be merged tomorrow AM.  @KevinCorcoran WDYT?

project.clj
KevinCorcoran: I would be tempted to use the newest version.  @pcarlisle thoughts?

pcarlisle: I can't see any reason not to, and it's faster

justinstoller:
I've updated cljfmt to the latest release (now at 0.5.0). I think this is ready to go.

NB: depending on the outcome of https://github.com/puppetlabs/pl-clojure-style/pull/3 I will bump the version of the submodule, but I can do that in the PR that applies the updated styles (https://github.com/puppetlabs/trapperkeeper/pull/232).

justinstoller:
So if everyone's cool with this approach and the diff that comes out in #232. I can just merge the two PRs and bump the submodule here.

I think we're a thumbs up on this approach? But I don't want to rework the conflicts in #232 until we're ready to merge since any other PRs will cause conflicts with the style update.

cprice404:
I'm good with it if @pcarlisle and @KevinCorcoran are.  I figure this will probably require a few iterations before we settle on something, seems reasonable to go ahead and get the first iteration into place.

KevinCorcoran:
:see_no_evil:

PR #228, opened 2016-03-18
(TK-344) Explicitly specify all logback dependencies

cprice404:
This commit adds explicit dependencies on all of the logback
artifacts that we use, to ensure that they are all the same version.
This also ensures that if any downstream projects happen to bring
in a conflicting version, the conflict will be caught at build time.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -1,4 +1,5 @@
 (def ks-version "1.3.0")
+(def logback-version "1.1.3")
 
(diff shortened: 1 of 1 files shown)


camlow325:
:see_no_evil: @rlinehan - care to review?

rlinehan:
:+1: tested manually with puppet server, confirmed this fixes the problem.

PR #229, opened 2016-03-20
(TK-331) prioritize shutdowns

cprice404:
This PR does the following:

* Tighten up the LifeCycleTask schema, since the `task-function` is no longer optional.
* Prioritize shutdowns over other events by adding a second core.async channel for requesting shutdowns, and using a prioritized `alts!` in the worker loop when reading off of the channel.
* Add logic to clean off any queued messages from the lifecycle channels during shutdown, so that we can ensure that no threads get stuck in a blocking `put` when we close the channels.

### DIFF
--- a/src/puppetlabs/trapperkeeper/internal.clj
+++ b/src/puppetlabs/trapperkeeper/internal.clj
@@ -219,11 +219,15 @@
   "Initializes a 'worker' which will listen for lifecycle-related tasks and perform
   them on a background thread, to ensure that we aren't executing multiple lifecycle
   tasks simultaneously."
-  [lifecycle-chan :- (schema/protocol async-prot/Channel)
+  [lifecycle-channel :- (schema/protocol async-prot/Channel)
+   shutdown-channel :- (schema/protocol async-prot/Channel)
    shutdown-reason-promise :- IDeref]
+  (log/debug "Initializing lifecycle worker loop.")
   (async/go (try
               (loop []
-                (let [task (async/<! lifecycle-chan)]
+                (let [[task chan] (async/alts!
+                                   [shutdown-channel lifecycle-channel]
+                                   :priority true)]
                   (schema/validate LifeCycleTask task)
                   (let [{:keys [type task-function]} task]
                     (condp #(contains? %1 %2) type
@@ -232,6 +236,9 @@
                         (log/debug "Received shutdown command, shutting down services")
                         (task-function)
                         (log/debug "Service shutdown complete, exiting lifecycle worker loop")
+                        (async/close! shutdown-channel)
+                        (async/close! lifecycle-channel)
+                        ;; TODO: drain closed channels
                         :done)
 
                       #{:boot :restart}
@@ -243,6 +250,7 @@
 
                       (throw (IllegalStateException. "Unrecognized lifecycle task:" task))))))
               (catch Exception e
+                (log/debug e "Exception caught in lifecycle worker loop")
                 (deliver shutdown-reason-promise
                          {:cause :service-error
                           :error e})))))
@@ -382,14 +390,14 @@
    in reverse order (to account for dependency relationships)."
   [app-context :- (schema/atom a/TrapperkeeperAppContext)]
   (log/info "Beginning shutdown sequence")
-  (let [{:keys [ordered-services lifecycle-channel lifecycle-worker]} @app-context
+  (let [{:keys [ordered-services shutdown-channel lifecycle-worker]} @app-context
         shutdown-fn (fn [] (doseq [[service-id s] (reverse ordered-services)]
                              (try
                                (run-lifecycle-fn! app-context s/stop "stop" service-id s)
                                (catch Exception e
                                  (log/error e "Encountered error during shutdown sequence")))))]
-    (async/>!! lifecycle-channel {:type :shutdown
-                                  :task-function shutdown-fn})
+    (async/>!! shutdown-channel {:type :shutdown
+                                 :task-function shutdown-fn})
     ;; wait for the channel to send us the return value so we know it's done
     (async/<!! lifecycle-worker)
     (log/info "Finished shutdown sequence")))
@@ -478,6 +486,7 @@
    config-data-fn :- IFn]
   (let [shutdown-reason-promise (promise)
         lifecycle-channel (async/chan max-pending-lifecycle-events)
+        shutdown-channel (async/chan)
         ;; this is the application context for this app instance.  its keys
         ;; will be the service ids, and values will be maps that represent the
         ;; context for each individual service
@@ -485,8 +494,10 @@
                            :ordered-services []
                            :services-by-id {}
                            :lifecycle-channel lifecycle-channel
+                           :shutdown-channel shutdown-channel
                            :lifecycle-worker (initialize-lifecycle-worker
                                               lifecycle-channel
+                                              shutdown-channel
                                               shutdown-reason-promise)
                            :shutdown-reason-promise shutdown-reason-promise})
         service-refs (atom {})
--- a/src/puppetlabs/trapperkeeper/internal.clj
+++ b/src/puppetlabs/trapperkeeper/internal.clj
@@ -234,15 +234,14 @@
                       #{:shutdown}
                       (do
                         (log/debug "Received shutdown command, shutting down services")
+                        (async/close! shutdown-channel)
+                        (async/close! lifecycle-channel)
                         (task-function)
                         (log/debug "Service shutdown complete, exiting lifecycle worker loop")
-                        (async/close! shutdown-channel)
-                        ;; drain the shutdown channel to ensure that there are
+
+                        ;; drain the channels to ensure that there are
                         ;; no blocking puts
                         (while (async/poll! shutdown-channel))
-                        (async/close! lifecycle-channel)
-                        ;; drain the lifecycle channel to ensure that there are no
-                        ;; blocking puts
                         (while (async/poll! lifecycle-channel))
                         :done)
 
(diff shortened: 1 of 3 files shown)


src/puppetlabs/trapperkeeper/internal.clj
```diff
@@ -219,17 +219,31 @@
...
                   (schema/validate LifeCycleTask task)
                   (let [{:keys [type task-function]} task]
                     (condp #(contains? %1 %2) type
                       #{:shutdown}
                       (do
-                        (log/debug "Received shutdown command, lifecycle worker exiting.")
+                        (log/debug "Received shutdown command, shutting down services")
+                        (task-function)
+                        (log/debug "Service shutdown complete, exiting lifecycle worker loop")
+                        (async/close! shutdown-channel)
```
camlow325: Would it make sense to do the `async/close!` calls before the `task-function` one?  Since the `task-function` one could take a long time, I'm wondering if it would make sense to shut the incoming channels down right away since we already know once we've gotten into this code that we don't care to read anything else from them anyway.  Seems like that would allow any further putters onto the channel to be rejected more quickly?

camlow325:
Doesn't seem necessarily related to this work, but I'm wondering if we should also be doing an `async/close!` call on the lifecycle worker when we're done using it?  Maybe that happens implicitly when the `async/go` block that it was launched from completes - or at least when the worker falls out of scope - but I wasn't sure.

cprice404:
> Doesn't seem necessarily related to this work, but I'm wondering if we should also be doing an async/close! call on the lifecycle worker when we're done using it? Maybe that happens implicitly when the async/go block that it was launched from completes

Yes, it does.

src/puppetlabs/trapperkeeper/internal.clj
```diff
@@ -380,20 +395,19 @@
...
+                               (run-lifecycle-fn! app-context s/stop "stop" service-id s)
+                               (catch Exception e
+                                 (log/error e "Encountered error during shutdown sequence")))))]
+    (log/trace "Putting shutdown message on shutdown channel.")
+    (async/>!! shutdown-channel {:type :shutdown
+                                 :task-function shutdown-fn})
+    ;; wait for the channel to send us the return value so we know it's done
+    (log/trace "Waiting for response to shutdown message from lifecycle worker.")
+    (if (not (nil? (async/<!! lifecycle-worker)))
+      (log/info "Finished shutdown sequence"))))
```
camlow325: Would it be helpful to still write some kind of message out if the shutdown is aborted - like for those trying to coordinate the number of "putting shutodnw message" / "waiting for response" messages to the corresponding result.  Maybe just a `(log/trace "Shutdown already requested, aborting additional shutdown attempt")`.  Open to whatever wording would sound best to you.

camlow325:
:+1: Will wait to merge at least until #224 is merged and this is rebased on it.

I did some functional testing of this from the command line with Puppet Server built to use the work in this branch.  I hit CTRL-C to kill the process across multiple process instances.  Nothing obviously bad happened in that the process died with no weird output in the log but I didn't see evidence that the shutdown hook fired.  I think this has always been a bit flaky anyway, though, so not sure there's cause for concern.  I also tried mixing in `kill -HUP` and `kill -TERM` calls together in quick succession.  I saw the process shutdown cleanly in each case, with the presumably queued HUP calls after the TERM triggered the shutdown being ignored.  This all looks good to me.

Left a couple of suggestions that you might consider addressing before we'd merge this:

1) Might be worthwhile to move the close/poll logic during shutdown up before the call to `task-function`, in order to allow any duplicate shutdown attempts to bail out sooner?

2) Might be good to log a trace level message when a redundant shutdown attempt is aborted - just to match up trace messages during troubleshooting.

cprice404:
@camlow325 I think that both of your suggestions are reasonable, will make those changes after #224 is merged, when I'm doing the rebase.

cprice404:
Rebased and added a commit that implements both of @camlow325 's suggestions.

src/puppetlabs/trapperkeeper/internal.clj
```diff
@@ -219,11 +219,15 @@
   "Initializes a 'worker' which will listen for lifecycle-related tasks and perform
   them on a background thread, to ensure that we aren't executing multiple lifecycle
   tasks simultaneously."
-  [lifecycle-chan :- (schema/protocol async-prot/Channel)
+  [lifecycle-channel :- (schema/protocol async-prot/Channel)
+   shutdown-channel :- (schema/protocol async-prot/Channel)
    shutdown-reason-promise :- IDeref]
+  (log/debug "Initializing lifecycle worker loop.")
   (async/go (try
               (loop []
```
KevinCorcoran: This comment is not about changes in this PR, but there's `async/go-loop` for this sort of thing.  Any reason that couldn't be used here?

cprice404: Because of the `try`?

src/puppetlabs/trapperkeeper/internal.clj
```diff
@@ -219,19 +219,30 @@
...
                         (log/debug "Received shutdown command, shutting down services")
+                        (async/close! shutdown-channel)
+                        (async/close! lifecycle-channel)
                         (task-function)
                         (log/debug "Service shutdown complete, exiting lifecycle worker loop")
+
+                        ;; drain the channels to ensure that there are
+                        ;; no blocking puts
+                        (while (async/poll! shutdown-channel))
+                        (while (async/poll! lifecycle-channel))
```
KevinCorcoran: Am I correct in thinking that these `poll!` calls are only for any threads which were already blocked or parked on a put when the channels were closed above?

src/puppetlabs/trapperkeeper/internal.clj
KevinCorcoran: Couldn't it be ...
```clj
(try
  (async/go-loop []
    ...
```
I know that's not exactly the same but it seems functionally equivalent.  I don't think that's an important change to make, just curious.

mullr: I don't think that's the same.. a go block happens on the core.async thread pool, so I wouldn't expect exception to bubble out.

src/puppetlabs/trapperkeeper/internal.clj
```diff
@@ -382,17 +394,22 @@
...
+                                 :task-function shutdown-fn})
     ;; wait for the channel to send us the return value so we know it's done
-    (async/<!! lifecycle-worker)
-    (log/info "Finished shutdown sequence")))
+    (log/trace "Waiting for response to shutdown message from lifecycle worker.")
+    (if (not (nil? (async/<!! lifecycle-worker)))
+      (log/info "Finished shutdown sequence")
+      ;; else, there was already a shutdown in progress and the second one was ignored
+      (log/trace (str "Response from lifecycle worker indicates shutdown already in progress,"
+                      "ignoring additional shutdown attempt.")))))
```
KevinCorcoran: The "else" case here is specifically because the `lifecycle-worker` channel is already closed, right?  If so, would you mind updating the comment and/or log message to clarify that?

src/puppetlabs/trapperkeeper/internal.clj
KevinCorcoran: Ah, right, I guess that would be wrapping whatever magic `go` does with error handling, which is not what we want here.

KevinCorcoran:
:see_no_evil:   Just a couple unanswered questions at this point.

src/puppetlabs/trapperkeeper/internal.clj
cprice404: That is correct, and yeah I can try to make it more clear.

src/puppetlabs/trapperkeeper/internal.clj
cprice404: correct

KevinCorcoran: Cool, I might add "since these channels were just closed above" to the end of that comment, or something like that.

src/puppetlabs/trapperkeeper/internal.clj
cprice404: In thinking about this further, though, I think maybe I need to move that try inside of the loop for a different reason (because if the loop exits then there will be no one left to try to respond to the shutdown message; this was a change in behavior introduced in my previous PR), so I'll see about fixing that in another commit when I update the comment that you mentioned.

cprice404:
Added a commit that improves the comment that @KevinCorcoran referenced, and cleans up the try/catch logic inside of the core.async loop a bit.

camlow325:
:+1:

src/puppetlabs/trapperkeeper/internal.clj
```diff
@@ -219,33 +219,60 @@
...
+                                       (:type msg)))
+              (ks/while-let [msg (async/poll! lifecycle-channel)]
+                            (log/debug "Shutdown in progress, ignoring message on main lifecycle channel:"
+                                       (:type msg)))
+
+              (task-function)
+
+              (log/debug "Service shutdown complete, exiting lifecycle worker loop")
               (catch Exception e
+                (log/debug e "Exception caught during shutdown!")))
```
KevinCorcoran: I think I prefer `error` or `warn` instead of `debug` here.

KevinCorcoran:
Should there be some test coverage for the changes in the last 2 commits?

Otherwise, :+1:

camlow325:
I think the last two comments @KevinCorcoran had listed about changing one of the shutdown log messages to a `log/error` instead of `log/debug` and adding a test to exercise the exception handling are valid.  I don't believe, however, that any existing production code paths would get us into the exception handling code for either of these cases because the `task-function`s which are executed for :boot, :shutdown, and :restart all have their own inner try/catch blocks that end up eating any lower-level exceptions that would be thrown.  So effectively, this extra error handling code added for the `async/go-loop` should never be reached in production but seems good to include anyway for robustness.

I'm going to go ahead and merge this one.  If it would be useful, I'd be good with having a follow-up PR submitted later to add test coverage and change the message from a debug->error.

PR #231, opened 2016-03-29
(TK-347) Add support for multiple bootstrap files

jpinsonault:
This PR adds support for bootstrap configurations to be split up between
multiple .cfg files.

All the .cfg files are essentially concatenated together before tk loads any of
the services in them, so that it appears as one combined bootstrap file. Nothing
changes with how TK loads the services.

Multiple bootstrap files are specified on the command line in the same way as
conf are: a comma separated list of files and directories. Because of this,
these changes are backward compatible with specifying a single file.

For example, the --bootstrap-config command line option might look like so:
`--bootstrap-config /path/to/bootstrap.cfg,/folder/of/bootstrap/files,./another.cfg`


jpinsonault:
I'd recommend looking at the two commits separately, as the second one is entirely updating functions to use schema.

Also I'll need to update the docs before this is merged, but I figured people could look at the implementation

src/puppetlabs/trapperkeeper/bootstrap.clj
```diff
@@ -58,42 +60,45 @@
...
+   - If the path doesn't lead to a file or directory, attempt to load a file from
+   a URI (for files in jars)"
+  [config-path :- schema/Str]
+  (cond
+    (fs/file? config-path) [(fs/file config-path)]
+    (fs/directory? config-path) (fs/glob (fs/file config-path "*.cfg"))
+    :else (try
+            [(io/input-stream (URI. config-path))]
+            (catch Exception ignored
+              ;; If that didn't work either, we can't read it.  Don't wrap and
```
fpringvaldsen: Might want to remove the `If that didn't work either` bit

jpinsonault:
Rebased to fix the failing commit

fpringvaldsen:
:+1:

documentation/Command-Line-Arguments.md
```diff
@@ -11,9 +11,24 @@ Note that if you absolutely need control over the command line argument processi
...
+
+Multiple bootstrap files are specified by giving the `--bootstrap-config` command line option a comma separated list of files and directories. For example:
+```
+--bootstrap-config ./first/path,/etc/second/path,./a/single/file.cfg
+```
+
+Each item in the list of paths can be one of:
+* A path to a single config file
+* A path to a directory of config files. Only files ending in .cfg will be used
+* A path a file inside of a jar. E.g. `jar:file:///usr/bin/myjar.jar!/bootstrap.cfg`
```
fpringvaldsen: s/a file/to a file

jpinsonault:
Addressed your feedback @fpringvaldsen

documentation/Command-Line-Arguments.md
```diff
@@ -11,9 +11,24 @@ Note that if you absolutely need control over the command line argument processi
 Trapperkeeper supports three command-line arguments:
 
 * `--config/-c`: The path to the configuration file or directory. This option is used to initialize the configuration service. This argument is optional; if not specified, Trapperkeeper will act as if you had given it an empty configuration file.
-* `--bootstrap-config/-b`: This argument is optional; if specified, the value should be a path to a bootstrap configuration file that Trapperkeeper will use (instead of looking for `bootstrap.cfg` in the current working directory or on the classpath)
+* `--bootstrap-config/-b`: This argument is optional; if specified, the value should be a path to a bootstrap configuration file, or a comma separated list of files and directories ([see below](#multiple-booststrap-files)) that Trapperkeeper will use (instead of looking for `bootstrap.cfg` in the current working directory or on the classpath)
 * `--debug/-d`: This option is not required; it's a flag, so it will evaluate to a boolean.  If `true`, sets the logging level to DEBUG, and also sets the `:debug` key in the configuration map provided by the configuration-service.
 
+### Multiple booststrap files
```
camlow325: s/booststrap/bootstrap here and at the other places that link to this.

documentation/Command-Line-Arguments.md
```diff
@@ -11,9 +11,24 @@ Note that if you absolutely need control over the command line argument processi
 Trapperkeeper supports three command-line arguments:
 
 * `--config/-c`: The path to the configuration file or directory. This option is used to initialize the configuration service. This argument is optional; if not specified, Trapperkeeper will act as if you had given it an empty configuration file.
```
camlow325: Unrelated to this PR since this functionality was pre-existing, but might be good to update this line while we're in here to indicate that a comma separated list of files / directories can be provided for this option also.

src/puppetlabs/trapperkeeper/bootstrap.clj
```diff
@@ -5,38 +5,40 @@
...
 
+;; Schemas
+(def BootstrapLine {:namespace schema/Str :service-name schema/Str})
+(def BootstrapFiles [(schema/protocol io/IOFactory)])
+
+;; Constants
 (def bootstrap-config-file-name "bootstrap.cfg")
 
-(defn- parse-bootstrap-line!
+(schema/defn ^:private parse-bootstrap-line! :- BootstrapLine
```
camlow325: Not sure if it's better to mark these `^:private` or just create a `private` "pragma" section for these, like:

````clj
;;;;;;;;;;;;;;;;;;;;;;;
;;; Private
````
@KevinCorcoran - do you know if we have a standard / preference for what we're doing on these now?

File: src/puppetlabs/trapperkeeper/internal.clj [hunk 6]
```diff
@@ -135,16 +136,14 @@
...
   "Parses the command-line arguments using `puppetlabs.kitchensink.core/cli!`.
   Hard-codes the command-line arguments expected by trapperkeeper to be:
       --debug
       --bootstrap-config <bootstrap file>
       --config <.ini file or directory>
       --plugins <plugins directory>"
-  [cli-args]
-  {:pre  [(sequential? cli-args)]
-   :post [(map? %)]}
+  [cli-args :- [(schema/maybe schema/Str)]]
```
camlow325: Could this just be...

```clj
[cli-args :- [schema/Str]]
```
?

Wasn't sure if it was expected / possible for one of the arguments to be nil.

test/puppetlabs/trapperkeeper/bootstrap_test.clj
```diff
@@ -154,21 +156,70 @@ puppetlabs.trapperkeeper.examples.bootstrapping.test-services/hello-world-servic
...
                                  (StringReader.)
-                                 parse-bootstrap-config!
+                                 vector
+                                 parse-bootstrap-configs!
                                  (map service-map))]
       (is (= (count service-maps) 2))
       (is (contains? (first service-maps) :HelloWorldService))
       (is (contains? (second service-maps) :TestService)))))
 
+(deftest multiple-bootstrap-files
```
camlow325: I think it would be good to have a test where more than one service is present in an individual bootstrap file.  Maybe a combo test that does multiple files in one directory where each file has more than one service present in the bootstrap file.

test/puppetlabs/trapperkeeper/bootstrap_test.clj
```diff
@@ -154,21 +156,70 @@ puppetlabs.trapperkeeper.examples.bootstrapping.test-services/hello-world-servic
...
+            hello-world-svc (get-service app :HelloWorldService)]
+        (is (logged?
+              ; We can't know what order it will find the files on disk, so just
+              ; look for a partial match with the path we gave TK.
+              (re-pattern (format "Loading bootstrap configs:\n%s"
+                                  (fs/absolute bootstrap-path)))
+              :debug))
+        (is (= (test-fn test-svc) :cli))
+        (is (= (hello-world hello-world-svc) "hello world")))))
+  (testing "A path containing both a file and a folder can be specified on the command line"
```
camlow325: I think it would be good to have a variation of this where the services found from the directory are different than the services found from the files but that all of the expected services from all locations were still loaded properly.

The loading of bootstraps from the /one directory vs. /one/bootstrap_one.cfg would seem useful to preserve as a separate test which proves that duplicate entries across files can be present without causing any problems.

test/puppetlabs/trapperkeeper/bootstrap_test.clj
```diff
@@ -154,21 +156,70 @@ puppetlabs.trapperkeeper.examples.bootstrapping.test-services/hello-world-servic
...
+        (is (= (test-fn test-svc) :cli))
+        (is (= (hello-world hello-world-svc) "hello world"))))))
+
 (deftest bootstrap-path-with-spaces
   (testing "Ensure that a bootstrap config can be loaded with a path that contains spaces"
     (with-test-logging
-      (let [app             (bootstrap-with-empty-config
-                               ["--bootstrap-config" "./dev-resources/bootstrapping/cli/path with spaces/bootstrap.cfg"])
-            test-svc        (get-service app :TestService)
+      (let [bootstrap-path "./dev-resources/bootstrapping/cli/path with spaces/bootstrap.cfg"
```
camlow325: It might also be good to have a variation of this test that has two bootstrap files where each file has spaces in the name - just to confirm that that split we do on file names doesn't adversely affect the ability to use any individual files that had spaces in the name.

camlow325:
Overall, big :+1:  Very nice cleanup of the code around the additions of schema.

I played around a bit with this by using it from Puppet Server with two bootstrap configs - one with just the ca service and one with all of the other services.  All of the services loaded and worked as I expected.

I left a few comments about some additional tests that I think it would be good to consider adding.  The most useful of those I'd think would be to have a test validate that all services are loaded fine when you have multiple bootstrap files with multiple service references in them.

cprice404:
@jpinsonault thanks a bunch for all of the schema cleanup, that's awesome.

One thing I'd recommend doing prior to this being merged, though, is to do a `lein install` of this on your machine, and then grabbing the master branch of puppetserver, pointing it to this snapshot version of TK, and running those tests.  Occasionally the downstream apps end up tickling code paths that would be affected by schemas that don't get caught in the TK tests.

documentation/Command-Line-Arguments.md
jpinsonault: Yeah, good idea

documentation/Command-Line-Arguments.md
jpinsonault: booststrapping is like a rocket fueled version of bootstrapping

src/puppetlabs/trapperkeeper/internal.clj
jpinsonault: So this one is a little weird. This line: https://github.com/puppetlabs/trapperkeeper/blob/master/test/puppetlabs/trapperkeeper/bootstrap_test.clj#L93

Passes nil into `parse-cli-args!`, to test that a "Missing required argument for --bootstrap-config" error is raised. I couldn't find a way to pass, say, an empty string in or something like that get the same behavior. Honestly I'm not sure if that test is worth having though, it seems to only be testing the `puppetlabs.kitchensink/cli!` function.

I can't think of any legitimate reason to be passing nils into `parse-cli-args!`

test/puppetlabs/trapperkeeper/bootstrap_test.clj
jpinsonault: This test loads `one/` and `two/bootstrap_two.cfg`, so they are different services.

How does TK handle duplicate entries currently? I imagine it has the same behavior described in TK-211 where if two services implement the same protocol it'll just load the second one and ignore the first. Maybe allowing duplicate entries (in addition to disallowing duplicate service protocol implementations) should be done as part of TK-211

test/puppetlabs/trapperkeeper/bootstrap_test.clj
jpinsonault: Yeah, good idea

test/puppetlabs/trapperkeeper/bootstrap_test.clj
jpinsonault: Do you think we need any tests with just one line in the bootstrap file? Or should I just switch this test to using more than one service in each file?

src/puppetlabs/trapperkeeper/internal.clj
camlow325: Seems like it would be better to have the schema reflect the real world data that we'd expect to see rather than loosening the schema to satisfy the needs of a unit test.

For that particular unit test, it seems like we're just wanting to validate that if no argument is provided to the right of "--bootstrap-config" that an appropriate error occurs.  Another way to do that might be to change the implementation of the `bootstrap-with-empty-config` function to put the `--config ...` option at the front of the argument list, like this:

```clj
(->> other-args
     (concat ["--config" empty-config])
     (vec)
```
... so that the test can just pass a value like ... 

```
(bootstrap-with-empty-config ["--bootstrap-config"])
```

... without requiring a bogus / non-real world `nil` argument to be appended to the end?

documentation/Command-Line-Arguments.md
camlow325: :rocket:

test/puppetlabs/trapperkeeper/bootstrap_test.clj
camlow325: I'd be good either way on that.  If you think there's some unique code that could be executed for a single line vs. multiple lines, maybe a separate test is warranted.  If not, though, multiple lines in each seems fine.

test/puppetlabs/trapperkeeper/bootstrap_test.clj
camlow325: Ah, yep, those are different services.  Mixed up my booostraps.  :smile: 

My recollection was that the current implementation would just load a single instance of the service even if the same namespace were to appear multiple times in the file.  I tried doing a split bootstrap install where I duplicated the namespace in each and it seemed to work fine with changes on your branch.  Assuming you agree it does / should as well, I was just thinking you could add a test onto this PR which demonstrates that.  If it turns out something is broken and you think it would be better to fix that as part of the work to disallow different service namespaces that implement the same protocol in [TK-211](https://tickets.puppetlabs.com/browse/TK-211), I'm fine deferring out the addition of a duplicates test in this PR.  Make sense?

src/puppetlabs/trapperkeeper/internal.clj
jpinsonault: Sounds good to me

test/puppetlabs/trapperkeeper/bootstrap_test.clj
jpinsonault: Cool, yeah there's nothing about it that would treat one line differently. A test with empty bootstrap configs in the same folder as normal ones would probably be a good idea though

test/puppetlabs/trapperkeeper/bootstrap_test.clj
jpinsonault: Yeah they're definitely being allowed (maybe by accident) right now, and just talked to Chris and he agrees that duplicate entries should be allowed. I'll add a test for duplicates

jpinsonault:
@camlow325 I think I addressed all your feedback

Also I tried running the puppetserver tests and there are some problems it looks like. Investigating

jpinsonault:
puppetserver unit tests were failing because I renamed `parse-bootstrap-config!`
Added that function back in, puppetserver tests are passing locally for me

camlow325:
@cprice404 Do you consider `parse-bootstrap-config!` to be "public API" in Trapperkeeper that we'd need to preserve?  If not, might be better to just pull it out and change the Puppet Server tests later to use the replacement.

cprice404:
Yeah, I think I consider that to be public API.

test/puppetlabs/trapperkeeper/bootstrap_test.clj
```diff
@@ -180,3 +269,24 @@ puppetlabs.trapperkeeper.examples.bootstrapping.test-services/foo-test-service ;
       ;; just test that this bootstrap config file can be read successfully
       ;; (ie, this does not throw an exception)
       (bootstrap-with-empty-config ["--bootstrap-config" bootstrap-url]))))
+
+(deftest duplicate-service-definitions
```
camlow325: I don't understand what this test is doing.  It doesn't appear to assert anything.  Should this test - and the associated `duplicate_services.cfg` - be part of the work that you're doing for [TK-211](https://tickets.puppetlabs.com/browse/TK-211)?

camlow325:
:+1: other than the remaining question I had about the `duplicate-service-definitions` test.

test/puppetlabs/trapperkeeper/bootstrap_test.clj
jpinsonault: Ah, yeah, sorry. I'd been working on that other ticket and forgot to switch branches before switching back to this ticket and did a git add -p. Missed this one.

camlow325: I saw that you removed the test in a later commit.  Would seem good to remove the `duplicate_services.cfg` file from the PR as well since nothing in this PR uses it.

jpinsonault: (facepalm) Yes, yes it would. Updated

camlow325:
:+1:

src/puppetlabs/trapperkeeper/bootstrap.clj
KevinCorcoran: I think the pragmas are more commonplace among our other projects but I wouldn't call it a standard or defined best practice.

PR #233, opened 2016-04-07
(TK-211) Disallow duplicate protocol implementations

jpinsonault:
This PR extends TK to throw an error during startup if two bootstrap
entries implement the same service protocol. Previously TK would startup the
last entry it found that implemented a protocol and silently ignore the rest.

This PR also contains additions to the bootstrap namespace to give more
information when a bootstrap related error is raised. Since bootstrap entries
can now be split up between multiple files, the error messages give the
location of the bootstrap files and the line number where the offending
service entries were included.

This functionality is implemented by passing around a list of the bootstrap
entries annotated with the file and line number they came from.

This PR also makes some changes to maintain the ability to have duplicate
bootstrap entries, since TK might now see them as two services that implement
the same protocol, and error.

### DIFF
--- a/dev-resources/bootstrapping/cli/bootstrap_with_comments.cfg
+++ b/dev-resources/bootstrapping/cli/bootstrap_with_comments.cfg
@@ -1,0 +1,5 @@
+# commented out line
+puppetlabs.trapperkeeper.examples.bootstrapping.test-services/hello-world-service # comment
+; another commented out line
+ ;puppetlabs.trapperkeeper.examples.bootstrapping.test-services/foo-test-service
+puppetlabs.trapperkeeper.examples.bootstrapping.test-services/foo-test-service ; comment
--- a/src/puppetlabs/trapperkeeper/bootstrap.clj
+++ b/src/puppetlabs/trapperkeeper/bootstrap.clj
@@ -82,7 +82,7 @@
     (map str (fs/glob (fs/file config-path "*.cfg")))
     [config-path]))
 
-(schema/defn ^:private config-from-cli :- (schema/maybe [schema/Str])
+(schema/defn ^:private config-from-cli :- [schema/Str]
   "Given the data from the command-line (parsed via `core/parse-cli-args!`),
   check to see if the caller explicitly specified the location of one or more
   bootstrap config files.  If so, return an object that can be read via
@@ -99,7 +99,7 @@
                            (string/join "\n" config-files)))
         config-files))))
 
-(schema/defn ^:private config-from-cwd :- (schema/maybe [schema/Str])
+(schema/defn ^:private config-from-cwd :- [schema/Str]
   "Check to see if there is a bootstrap config file in the current working
   directory;  if so, return it."
   []
@@ -188,25 +188,23 @@
   (->> coll
        (group-by f)
        ; filter out map values with only 1 item
-       (remove #(= 1 (count (val %))))))
+       (remove #(= 1 (count (val %))))
+       (into {})))
 
 (schema/defn duplicate-protocol-error :- IllegalArgumentException
   "Returns an IllegalArgumentException describing what services implement
    the same protocol, including the line number and file the bootstrap entries
    were found"
-  [protocol-id :- schema/Keyword
-   duplicate-services :- [(schema/protocol services/ServiceDefinition)]
+  [duplicate-services :- {schema/Keyword [(schema/protocol services/ServiceDefinition)]}
    service->entry-map :- {(schema/protocol services/ServiceDefinition) AnnotatedBootstrapEntry}]
   (let [make-error-message (fn [service]
                              (let [entry (get service->entry-map service)]
-                               (format "%s:%s\n%s"
-                                       (:bootstrap-file entry)
-                                       (:line-number entry)
-                                       (:entry entry))))]
-    (IllegalArgumentException.
-     (format (str "Duplicate implementations found for service protocol '%s':\n%s")
-             protocol-id
-             (string/join "\n" (map make-error-message duplicate-services))))))
+                               (format "%s:%s\n%s" (:bootstrap-file entry) (:line-number entry) (:entry entry))))]
+    (let [error-messages (for [[protocol-id services] duplicate-services]
+                           (format (str "Duplicate implementations found for service protocol '%s':\n%s")
+                                   protocol-id
+                                   (string/join "\n" (map make-error-message services))))]
+      (IllegalArgumentException. (string/join "\n" error-messages)))))
 
 (schema/defn check-duplicate-service-implementations!
   "Throws an exception if two services implement the same service protocol"
@@ -218,8 +216,9 @@
   (let [service->entry-map (zipmap services bootstrap-entries)]
     ; Find duplicates base on the service id returned by calling service-def-id
     ; on each service
-    (if-let [duplicate (first (find-duplicates services services/service-def-id))]
-      (throw (duplicate-protocol-error (key duplicate) (val duplicate) service->entry-map)))))
+    (let [duplicates (find-duplicates services services/service-def-id)]
+      (when (not (empty? duplicates))
+        (throw (duplicate-protocol-error duplicates service->entry-map))))))
 
 (schema/defn bootstrap-error :- IllegalArgumentException
   "Returns an IllegalArgumentException meant to wrap other errors relating to
@@ -280,13 +279,16 @@
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 
 (schema/defn parse-bootstrap-configs! :- [(schema/protocol services/ServiceDefinition)]
-  [configs :- [schema/Str]]
   "Parse multiple trapperkeeper bootstrap configuration files and return the
   service graph that is the result of merging the graphs of all of the
   services specified in the configuration files."
-  ; We remove the duplicate entries to allow the user to have duplicate entries in their
-  ; bootstrap files. If we didn't remove them, it would look like two services were trying
-  ; to implement the same protocol when we check for duplicate service implementations
+  [configs :- [schema/Str]]
+  ; We remove the duplicate entries (the exact same namespace and name) to allow
+  ; the user to have duplicate entries in their bootstrap files. If we didn't
+  ; remove them, it would look like two services were trying to implement the
+  ; same protocol when we check for duplicate service implementations. We want
+  ; to allow entries with the same exact name in order to support workflows
+  ; where users are preparing to upgrade and might copy an entry to another file.
   (let [bootstrap-entries (remove-duplicate-entries (get-annotated-bootstrap-entries configs))]
     (when (empty? bootstrap-entries)
       (throw (Exception. "Empty bootstrap config file")))
--- a/src/puppetlabs/trapperkeeper/bootstrap.clj
+++ b/src/puppetlabs/trapperkeeper/bootstrap.clj
@@ -218,7 +218,7 @@
   (for [{:keys [bootstrap-file line-number entry]} bootstrap-entries]
     (try+
       (let [{:keys [namespace service-name]} (parse-bootstrap-line! entry)]
-        (resolve-service! namespace service-name ))
+        (resolve-service! namespace service-name))
       ; Catch and re-throw as java exception
       (catch [:type ::bootstrap-parse-error] {:keys [message]}
         (throw (bootstrap-error entry bootstrap-file line-number message)))
--- a/src/puppetlabs/trapperkeeper/bootstrap.clj
+++ b/src/puppetlabs/trapperkeeper/bootstrap.clj
@@ -204,9 +204,9 @@
                                        (:line-number entry)
                                        (:entry entry))))]
     (IllegalArgumentException.
-      (format (str "Duplicate implementations found for service protocol '%s':\n%s")
-              protocol-id
-              (string/join "\n" (map make-error-message duplicate-services))))))
+     (format (str "Duplicate implementations found for service protocol '%s':\n%s")
+             protocol-id
+             (string/join "\n" (map make-error-message duplicate-services))))))
 
 (schema/defn check-duplicate-service-implementations!
   "Throws an exception if two services implement the same service protocol"
@@ -230,9 +230,9 @@
    line-number :- schema/Int
    original-message :- schema/Str]
   (IllegalArgumentException.
-    (format (str "Problem loading service '%s' on line '%s' in bootstrap "
-                 "configuration file '%s':\n%s")
-            entry line-number bootstrap-file original-message)))
+   (format (str "Problem loading service '%s' on line '%s' in bootstrap "
+                "configuration file '%s':\n%s")
+           entry line-number bootstrap-file original-message)))
 
 (schema/defn resolve-services! :- [(schema/protocol services/ServiceDefinition)]
   "Resolves each bootstrap entry into an instance of a trapperkeeper
--- a/src/puppetlabs/trapperkeeper/bootstrap.clj
+++ b/src/puppetlabs/trapperkeeper/bootstrap.clj
@@ -13,7 +13,6 @@
 
 ;; Schemas
 (def ParsedBootstrapEntry {:namespace schema/Str :service-name schema/Str})
-(def BootstrapFiles [schema/Str])
 (def AnnotatedBootstrapEntry {:entry schema/Str
                               :bootstrap-file schema/Str
                               :line-number schema/Int})
@@ -62,7 +61,7 @@
       (string/replace #"(?:#|;).*$" "")
       (string/trim)))
 
-(schema/defn find-bootstraps-from-path :- BootstrapFiles
+(schema/defn find-bootstraps-from-path :- [schema/Str]
   "Given a path, return a list of .cfg files found there.
    - If the path leads directly to a file, return a list with a single item.
    - If the path leads to a directory, return a list of any .cfg files found there.
@@ -73,7 +72,7 @@
     (map str (fs/glob (fs/file config-path "*.cfg")))
     [config-path]))
 
-(schema/defn ^:private config-from-cli :- (schema/maybe BootstrapFiles)
+(schema/defn ^:private config-from-cli :- (schema/maybe [schema/Str])
   "Given the data from the command-line (parsed via `core/parse-cli-args!`),
   check to see if the caller explicitly specified the location of one or more
   bootstrap config files.  If so, return an object that can be read via
@@ -90,7 +89,7 @@
                            (string/join "\n" config-files)))
         config-files))))
 
-(schema/defn ^:private config-from-cwd :- (schema/maybe BootstrapFiles)
+(schema/defn ^:private config-from-cwd :- (schema/maybe [schema/Str])
   "Check to see if there is a bootstrap config file in the current working
   directory;  if so, return it."
   []
@@ -111,7 +110,7 @@
     (log/debug (str "Loading bootstrap config from classpath: '" classpath-config "'"))
     [(.getPath classpath-config)]))
 
-(schema/defn find-bootstrap-configs :- BootstrapFiles
+(schema/defn find-bootstrap-configs :- [schema/Str]
   "Get the bootstrap config files from:
     1. the file path specified on the command line, or
     2. the current working directory, or
@@ -151,6 +150,8 @@
     (line-seq (io/reader config-file))))
 
 (schema/defn get-annotated-bootstrap-entries :- [AnnotatedBootstrapEntry]
+  "Reads each bootstrap entry into a map with the line number and file the
+  entry is from. Returns a list of maps."
   [configs :- [schema/Str]]
   (for [config configs
         [line-number line-text] (indexed (map remove-comments (read-config config)))
@@ -160,7 +161,7 @@
      :entry line-text}))
 
 (defn find-duplicates
-  "Collects duplicates base on running f on each item.
+  "Collects duplicates base on running f on each item in coll.
    Returns a map where the keys will be the result of running f on each item,
    and the values will be lists of items that are duplicates of eachother"
   [coll f]
@@ -213,8 +214,11 @@
                  "configuration file '%s':\n%s")
             entry line-number bootstrap-file original-message)))
 
-(schema/defn resolve-services!
-  [bootstrap-entries]
+(schema/defn resolve-services! :- [(schema/protocol services/ServiceDefinition)]
+  "Resolves each bootstrap entry into an instance of a trapperkeeper
+  ServiceDefinition.
+  Throws an IllegalArgumentException if the service can't be resolved"
+  [bootstrap-entries :- [AnnotatedBootstrapEntry]]
   (for [{:keys [bootstrap-file line-number entry]} bootstrap-entries]
     (try+
       (let [{:keys [namespace service-name]} (parse-bootstrap-line! entry)]
@@ -253,6 +257,9 @@
 
 (schema/defn parse-bootstrap-configs! :- [(schema/protocol services/ServiceDefinition)]
   [configs :- [schema/Str]]
+  "Parse multiple trapperkeeper bootstrap configuration files and return the
+  service graph that is the result of merging the graphs of all of the
+  services specified in the configuration files."
   ; We remove the duplicate entries to allow the user to have duplicate entries in their
   ; bootstrap files. If we didn't remove them, it would look like two services were trying
   ; to implement the same protocol when we check for duplicate service implementations
--- a/src/puppetlabs/trapperkeeper/bootstrap.clj
+++ b/src/puppetlabs/trapperkeeper/bootstrap.clj
@@ -13,9 +13,9 @@
 
 ;; Schemas
 (def ParsedBootstrapEntry {:namespace schema/Str :service-name schema/Str})
-(def BootstrapFiles [(schema/protocol io/IOFactory)])
+(def BootstrapFiles [schema/Str])
 (def AnnotatedBootstrapEntry {:entry schema/Str
-                              :bootstrap-file (schema/protocol io/IOFactory)
+                              :bootstrap-file schema/Str
                               :line-number schema/Int})
 
 ;; Constants
@@ -69,20 +69,9 @@
    - If the path doesn't lead to a file or directory, attempt to load a file from
    a URI (for files in jars)"
   [config-path :- schema/Str]
-  (cond
-    (fs/file? config-path) [(fs/file config-path)]
-    (fs/directory? config-path) (fs/glob (fs/file config-path "*.cfg"))
-    :else (try
-            [(io/input-stream (URI. config-path))]
-            (catch Exception ignored
-              ;; If we can't find a file or a directory, and loading the URI
-              ;; fails, we give up and throw an exception.
-              ;; Don't wrap and re-throw here, as `ignored` may be misleading
-              ;; (in the case of a normal path, it is useless - there is no
-              ;; reason to mess with URIs)
-              (throw (IllegalArgumentException.
-                      (str "Specified bootstrap config file does not exist: '"
-                           config-path "'")))))))
+  (if (fs/directory? config-path)
+    (map str (fs/glob (fs/file config-path "*.cfg")))
+    [config-path]))
 
 (schema/defn ^:private config-from-cli :- (schema/maybe BootstrapFiles)
   "Given the data from the command-line (parsed via `core/parse-cli-args!`),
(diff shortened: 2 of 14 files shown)


src/puppetlabs/trapperkeeper/bootstrap.clj
```diff
@@ -132,26 +136,102 @@
...
+
+(schema/defn remove-duplicate-entries :- [AnnotatedBootstrapEntry]
+  "Removes duplicate entries by only looking at the :entry key in each entry map.
+   This way, duplicate entries in different files and on different lines are removed"
+  [entries :- [AnnotatedBootstrapEntry]]
+  ; Construct a [k v] pair where the key is the text of the bootstrap entry, and the
+  ; value is the original AnnotatedBootstrapEntry map. By shoving them into a map,
+  ; duplicate keys are ignored. We pick out the values with vals, and get back our
+  ; original list minus the duplicates.
+  (vals (into {} (for [entry entries] [(:entry entry) entry]))))
```
jpinsonault: I don't like magical one liners, but this was the easiest to read method I could come up with. Happy to take suggestions.

src/puppetlabs/trapperkeeper/bootstrap.clj
```diff
@@ -27,10 +31,10 @@
                                            #"^([a-zA-Z0-9\.\-]+)/([a-zA-Z0-9\.\-]+)$"
                                            line)]
     {:namespace namespace :service-name service-name}
-    (throw (IllegalArgumentException.
-            (str "Invalid line in bootstrap config file:\n\n\t"
-                 line
-                 "\n\nAll lines must be of the form: '<namespace>/<service-fn-name>'.")))))
+    (throw+ {:type :illegal-argument
```
cprice404: this is a nitpick, but I might use a type that was more specific here; maybe something like `::bootstrap-parse-error`.  The only reason we used a generic type (`IllegalArgumentException`) above is because we didn't want to have to create a new Java Exception class.

src/puppetlabs/trapperkeeper/bootstrap.clj
```diff
@@ -8,16 +8,20 @@
...
-            [me.raynes.fs :as fs]))
+            [me.raynes.fs :as fs]
+            [slingshot.slingshot :refer [try+ throw+]]))
 
 ;; Schemas
-(def BootstrapLine {:namespace schema/Str :service-name schema/Str})
+(def ParsedBootstrapEntry {:namespace schema/Str :service-name schema/Str})
 (def BootstrapFiles [(schema/protocol io/IOFactory)])
+(def AnnotatedBootstrapEntry {:entry schema/Str
+                    :bootstrap-file schema/Str
```
fpringvaldsen: Nitpick but the indentation on this is wrong

src/puppetlabs/trapperkeeper/bootstrap.clj
fpringvaldsen: :+1: on this

src/puppetlabs/trapperkeeper/bootstrap.clj
```diff
@@ -132,26 +136,102 @@
...
-      (let [{:keys [namespace service-name]} (parse-bootstrap-line! line)]
-        (resolve-service! namespace service-name)))))
+  ; We remove the duplicate entries to allow the user to have duplicate entries in their
+  ; bootstrap files. If we didn't remove them, it would look like two services were trying
+  ; to implement the same protocol when we check for duplicate service implementations
+  (let [bootstrap-entries (remove-duplicate-entries (get-annotated-bootstrap-entries configs))]
+    (when (empty? bootstrap-entries)
+      (throw (Exception. "Empty bootstrap config file")))
+    (let [resolved-services (resolve-services! bootstrap-entries)
+          _ (check-duplicate-service-implementations resolved-services bootstrap-entries)]
```
fpringvaldsen: It's probably worth moving this down into the body of the `let` block

src/puppetlabs/trapperkeeper/internal.clj
```diff
@@ -49,15 +51,16 @@
   [service-def]
   {:post [(satisfies? s/ServiceDefinition %)]}
   (if-not (satisfies? s/ServiceDefinition service-def)
-    (throw (IllegalArgumentException.
-            (str "Invalid service definition; expected a service "
-                 "definition (created via `service` or `defservice`); "
-                 "found: " (pr-str service-def)))))
+    (throw+ {:type :illegal-argument
+            :message (str "Invalid service definition; expected a service "
```
fpringvaldsen: Indentation here is slightly off

test/puppetlabs/trapperkeeper/bootstrap_test.clj
```diff
@@ -270,17 +282,15 @@ puppetlabs.trapperkeeper.examples.bootstrapping.test-services/foo-test-service ;
...
-           (chain-files [bootstrap-one bootstrap-two])))))
-  (testing "chain-files removes empty files"
-    (let [bootstrap-one (StringReader. "one\ntwo")
-          bootstrap-two (StringReader. "")
-          bootstrap-three (StringReader. "three\nfour")]
-      (is (=
-           ["one", "two", "three", "four"]
-           (chain-files [bootstrap-one bootstrap-two bootstrap-three]))))))
+(deftest parse-bootstrap-config-throws
+  (testing "throws error with line number and file"
```
fpringvaldsen: Might be worth explaining that this is testing loading a service that doesn't exist.

src/puppetlabs/trapperkeeper/bootstrap.clj
```diff
@@ -132,26 +136,102 @@
...
+  (map vector (iterate inc 0) coll))
 
-   In the case of an empty file, nil would be returned by line-seq, so we remove
-   all the nils at the end"
-  [files :- [(schema/protocol io/IOFactory)]]
-  (remove nil? (flatten (map #(line-seq (io/reader %)) files))))
+(schema/defn get-annotated-bootstrap-entries :- [AnnotatedBootstrapEntry]
+  [configs :- [(schema/protocol io/IOFactory)]]
+  (flatten (for [config configs]
+             (for [[line-number line-text] (indexed (map remove-comments (line-seq (io/reader config))))
```
cprice404: If you are removing comments before you call `indexed` then the line numbers aren't going to be accurate, are they?

jpinsonault: Remove comments removes the comments from the string, but doesn't remove the whole string. Because comments are allowed after a bootstrap entry, it doesn't just does a substitution to remove the comment. The line below this one, `:when (not (empty? line-text))` takes care of the empty line case.

I just reused this `for` patten from what was already in the code. Might be able to make it more readable though

cprice404: ah, ok, that makes sense.

I think you can get rid of the nested `for` though, I think the binding form for `for` will support multiple vars.

jpinsonault: I think you're right, I'll do that

cprice404: It might be worth adding a couple of unit tests that just cover this function directly, including one that has a file with a commented line in it; that'd probably make it a little more obvious.

src/puppetlabs/trapperkeeper/bootstrap.clj
```diff
@@ -132,26 +136,102 @@
...
+                   :when (not (empty? line-text))]
+               {:bootstrap-file (str config)
```
cprice404: This usage doesn't necessarily seem to match up with the schema for the fn.  The schema says that `config` can be any `IOFactory`, and I think that things like `InputStream` would satisfy that.  Here, you're assuming that the value has a meaningful `toString`, and storing it in a key called `bootstrap-file`, which implies that it's expected to be a file path.  Maybe, based on the usages of this function,  you could actually tighten up the schema to something like String or File?  In which case this line would be fine.

src/puppetlabs/trapperkeeper/bootstrap.clj
```diff
@@ -8,16 +8,20 @@
...
+(def ParsedBootstrapEntry {:namespace schema/Str :service-name schema/Str})
 (def BootstrapFiles [(schema/protocol io/IOFactory)])
+(def AnnotatedBootstrapEntry {:entry schema/Str
+                    :bootstrap-file schema/Str
+                    :line-number schema/Int})
 
 ;; Constants
 (def bootstrap-config-file-name "bootstrap.cfg")
 
-(schema/defn ^:private parse-bootstrap-line! :- BootstrapLine
```
cprice404: I'm fine with removing the `^:private` metadata, but we should probably add some of the big public/private pragma blocks to this file to make it super obvious what we are considering as API.  Especially since you've added a bunch of little helper functions; I think there should probably only be two or three fns that are 'public' in this namespace.

src/puppetlabs/trapperkeeper/bootstrap.clj
```diff
@@ -132,26 +136,102 @@
...
                  "command line argument, current working directory, or on classpath")))))
 
-(schema/defn chain-files :- [schema/Str]
-  "Takes a list of files, reads all their lines in, and returns a flattened seq
-   of all their lines put together.
+(schema/defn indexed
+  "Returns seq of [index, item] pairs
+  [:a :b :c] -> ([0 :a] [1 :b] [2 :c])"
+  [coll]
+  (map vector (iterate inc 0) coll))
```
cprice404: this might be more readable as:

`(zipmap (range (count coll)) coll)`

or it might not.

src/puppetlabs/trapperkeeper/bootstrap.clj
```diff
@@ -132,26 +136,102 @@
...
+                               (format "%s:%s\n%s"
+                                       (:bootstrap-file entry)
+                                       (:line-number entry)
+                                       (:entry entry))))]
+    (IllegalArgumentException.
+      (format (str "Duplicate implementations found for service protocol '%s':\n%s")
+              protocol-id
+              (string/join "\n" (map make-error-message duplicate-services))))))
+
+(schema/defn check-duplicate-service-implementations
```
cprice404: probably name this with a trailing `!` to indicate that it has side effects (throws)

src/puppetlabs/trapperkeeper/bootstrap.clj
jpinsonault: Agreed that the schema is a little off. I guess the problem I see is that even if I tighten it up to be a string, somewhere else in the code we'll have to assume it has a meaningful `toString` when it gets passed in here. And at some point something will have to be printed in the error messages. I don't think we can make it `File` because if the bootstrap is in a jar file it'll be an `InputStream`.

src/puppetlabs/trapperkeeper/bootstrap.clj
```diff
@@ -132,26 +136,102 @@
...
+  "Collects duplicates base on running f on each item.
+   Returns a map where the keys will be the result of running f on each item,
+   and the values will be lists of items that are duplicates of eachother"
+  [coll f]
+  (->> coll
+       (group-by f)
+       ; filter by map value with only 1 item
+       (remove #(= 1 (count (val %))))))
+
+(schema/defn duplicate-protocol-error
```
cprice404: return value schema?

src/puppetlabs/trapperkeeper/bootstrap.clj
```diff
@@ -132,26 +136,102 @@
...
+
+  ; Zip up the services and bootstrap entries and construct a map out of them
+  ; to use as a lookup table below
+  (let [service->entry-map (zipmap services bootstrap-entries)]
+    ; Find duplicates base on the service id returned by calling service-def-id
+    ; on each service
+    (if-let [duplicate (first (find-duplicates services services/service-def-id))]
+      (throw (duplicate-protocol-error (key duplicate) (val duplicate) service->entry-map)))))
+
+(schema/defn bootstrap-error :- IllegalArgumentException
```
cprice404: docstring?

src/puppetlabs/trapperkeeper/bootstrap.clj
cprice404: We might want to log when we find duplicates, anyway, right?  So maybe we turn this into a `reduce` that does something like:

```clj
(reduce
   (fn [acc entry]
      (if (contains? acc entry)
         (do (log/warn "Dupe found with line numbers...") acc)
         (conj acc entry)))
    (set)
    entries)
```

src/puppetlabs/trapperkeeper/bootstrap.clj
cprice404: agree

test/puppetlabs/trapperkeeper/bootstrap_test.clj
cprice404: agree

cprice404:
Done with initial review; looking good.  In general I think it would be a good habit to try to add some more unit tests for a handful of the unit-testable functions that are not trivial to reason about.  So maybe a few of those, adding the pragmas to mark stuff as public / private, and the other small handful of comments I put up, and then I'll be +1.

src/puppetlabs/trapperkeeper/bootstrap.clj
jpinsonault: I think since zipmap returns a map we'll have ordering issues, but it looks like `(map vector (range) coll)` works

src/puppetlabs/trapperkeeper/bootstrap.clj
jpinsonault: Ah, sorry I didn't mean to leave this change in. Agreed on adding the comment blocks. It sounds like in general we don't like marking things `^:private` since you can't require them in test files. I'll reorganize the file. Should I just get rid of all the private metadata tags?

src/puppetlabs/trapperkeeper/bootstrap.clj
cprice404: If we are actually supporting something coming from inside of the jar file, I think what we'd want to use here is the resource URI, which would end up looking something like `classpath:org/my/package/resource.extension`.

I haven't stepped through the full stack of the code, but definitely at the time when you parse the CLI args, you know they are Strings, and if it's possible for those raw Strings to be passed directly into this fn, then I think it's safe to use `schema/Str` for the schema.

In the future we might end up needing to pre-process some of these strings to make them into an `IOFactory` before we call `io/reader` on them - e.g. if we need to load a resource out of a jar - but I think that would happen in a new function called from inside of here, and that fn might take an `IOFactory` as input and call `io/reader` on it or something?

src/puppetlabs/trapperkeeper/bootstrap.clj
cprice404: If it's easier you can do that re-organization in a follow-up PR.

Personally I'm fine with keeping the `^:private` metadata tags until they cause a problem for a given fn, but I'm also fine with removing them if we add the pragma stuff.

src/puppetlabs/trapperkeeper/bootstrap.clj
cprice404: cool

src/puppetlabs/trapperkeeper/bootstrap.clj
jpinsonault: agreed

src/puppetlabs/trapperkeeper/bootstrap.clj
jpinsonault: Yeah, that seems like a good way to do it

src/puppetlabs/trapperkeeper/bootstrap.clj
jpinsonault: Ah ok, I see what you're getting at. Agreed, taking in strings makes sense

jpinsonault:
@cprice404 I think I've addressed all the comments. I'd direct you to the `parse-bootstrap-configs uses strings` commit, since it's the biggest change. The one thing I'm not a fan of switching to passing the strings along to `parse-bootstrap-configs!` is that I would have liked to display the error messages with absolute file paths, but now we can't know if it's a URI or path. The functions for getting absolute paths don't seem to handle URIs well. That might not be a big deal though.

I added private/public comment blocks to the `bootstrap` namespace, I think I got the split correct.

jpinsonault:
Also I threw this into puppetserver and ran the puppetserver clojure tests successfully

src/puppetlabs/trapperkeeper/bootstrap.clj
```diff
@@ -132,30 +135,158 @@
...
+(schema/defn read-config :- [schema/Str]
+  "Opens a bootstrap file from either a path or URI, and returns each line from
+  the config. Throws an exception if the file can't be found or if the URI
+  can't be loaded"
+  [config-path :- schema/Str]
+  (let [config-file (if (fs/file? config-path)
+                      (fs/file config-path)
+                      (try
+                        (io/input-stream (URI. config-path))
+                        (catch Exception ignored
```
cprice404: Generally speaking it's Very Bad Juju to catch and swallow an exception type that is this generic.  Is there something more specific you could catch?

My guess is that if you pulled the `URI` constructor call out into a `let` block, you would be able to catch some kind of "malformed" exception for that, and then maybe catch an IOException for the call to `io/input-stream`?

If you head in that direction, would be good to add basic unit tests for `read-config` for both of those cases.

cprice404: Also I'd probably wrap when you re-throw, unless you can give an example of why not to do that.  It's generally a good idea to try to avoid loss of fidelity in error handling.

src/puppetlabs/trapperkeeper/bootstrap.clj
```diff
@@ -255,6 +265,10 @@
                        {:set #{} :ordered-entries []}
                        entries))))
 
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+;; Public
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
```
cprice404: pragmas look like they are in the right spots to me.

cprice404:
@jpinsonault lgtm; I'm +1 except for the error handling in the `read-config` fn.  In general I like the direction you took that function in, and I like how things are shaping up in terms of using the string schema everywhere until we get to the internals of that function.  Would just like to get rid of the super-generic catch-and-swallow.

jpinsonault:
@cprice404 updated the `read-config` error handling. There's a little duplicated code with the `(line-seq (io/reader...` appearing twice but it makes the exception catching much cleaner to read

jpinsonault:
Although... I guess there's a minute possibility of a race condition between testing that `config-path` is a file and reading it, so maybe I should catch a `FileNotFoundException` for it too

cprice404:
+1

src/puppetlabs/trapperkeeper/bootstrap.clj
```diff
@@ -58,29 +71,18 @@
...
-              ;; Don't wrap and re-throw here, as `ignored` may be misleading
-              ;; (in the case of a normal path, it is useless - there is no
-              ;; reason to mess with URIs)
-              (throw (IllegalArgumentException.
-                      (str "Specified bootstrap config file does not exist: '"
-                           config-path "'")))))))
-
-(schema/defn ^:private config-from-cli :- (schema/maybe BootstrapFiles)
+  (if (fs/directory? config-path)
+    (map str (fs/glob (fs/file config-path "*.cfg")))
```
camlow325: Does this implementation still handle the ability to load a file from a URI / file inside of a jar?

src/puppetlabs/trapperkeeper/bootstrap.clj
```diff
@@ -132,30 +135,168 @@
...
-   In the case of an empty file, nil would be returned by line-seq, so we remove
-   all the nils at the end"
-  [files :- [(schema/protocol io/IOFactory)]]
-  (remove nil? (flatten (map #(line-seq (io/reader %)) files))))
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+;; Public
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 
 (schema/defn parse-bootstrap-configs! :- [(schema/protocol services/ServiceDefinition)]
+  [configs :- [schema/Str]]
```
rlinehan: The arguments should go after the docstring.

src/puppetlabs/trapperkeeper/bootstrap.clj
```diff
@@ -132,30 +135,168 @@
...
+  [services :- [(schema/protocol services/ServiceDefinition)]
+   bootstrap-entries :- [AnnotatedBootstrapEntry]]
+
+  ; Zip up the services and bootstrap entries and construct a map out of them
+  ; to use as a lookup table below
+  (let [service->entry-map (zipmap services bootstrap-entries)]
+    ; Find duplicates base on the service id returned by calling service-def-id
+    ; on each service
+    (if-let [duplicate (first (find-duplicates services services/service-def-id))]
+      (throw (duplicate-protocol-error (key duplicate) (val duplicate) service->entry-map)))))
```
camlow325: So if we had duplicates for more than one protocol, would we only display the error for the first protocol where we found a duplicate?  That's probably fine - just seems like it might be nice to display all of the duplicates since `find-duplicates` would appear to actually do the work to locate them all.

src/puppetlabs/trapperkeeper/bootstrap.clj
```diff
@@ -97,27 +99,28 @@
...
-      (log/debug (str "Loading bootstrap config from current working directory: '"
-                      (.getAbsolutePath config-file) "'"))
-      [config-file])))
+      (let [config-file-path (.getAbsolutePath config-file)]
+        (log/debug (str "Loading bootstrap config from current working directory: '"
+                        config-file-path "'"))
+        [config-file-path]))))
 
-(schema/defn ^:private config-from-classpath :- (schema/maybe BootstrapFiles)
+(schema/defn ^:private config-from-classpath :- [(schema/maybe schema/Str)]
```
rlinehan: I don't think the `schema/maybe` is needed here. I'm pretty sure that in the way we are constructing URLs that `.getPath` won't return nil. I guess I'm not 100% positive though, so maybe safer to leave it in? But then also, if this is the schema here, then `find-bootstrap-configs` should also have this schema, since this could be one of the options that it returns.

src/puppetlabs/trapperkeeper/bootstrap.clj
```diff
@@ -97,27 +99,28 @@
                            (string/join "\n" config-files)))
         config-files))))
 
-(schema/defn ^:private config-from-cwd :- (schema/maybe BootstrapFiles)
+(schema/defn ^:private config-from-cwd :- (schema/maybe [schema/Str])
```
rlinehan: I think this should just be `[schema/Str]` - I'm pretty sure the `maybe` does nothing here?

```clojure
=> (schema/validate [schema/Str] nil)
nil
```

src/puppetlabs/trapperkeeper/bootstrap.clj
rlinehan: Agree with this - I think that it would be much more helpful to tell where both/all of the duplicates are than just the first.

src/puppetlabs/trapperkeeper/bootstrap.clj
camlow325: n/m, I see how this is done in `read-config` now.

jpinsonault: It does. That functionality is now inside of `read-config`

src/puppetlabs/trapperkeeper/bootstrap.clj
jpinsonault: Woah, I did not know that. I'm guessing it's because `nil` can be passed into functions that take seqs? Good to know

src/puppetlabs/trapperkeeper/bootstrap.clj
jpinsonault: I think this was supposed to be `(schema/maybe [schema/Str])`, which means your previous comment applies about it not being necessary. Should probably just be `[schema/Str]`

src/puppetlabs/trapperkeeper/bootstrap.clj
jpinsonault: Sounds good

src/puppetlabs/trapperkeeper/bootstrap.clj
```diff
@@ -132,30 +135,168 @@
...
-    (when (empty? lines) (throw (Exception. "Empty bootstrap config file")))
-    (for [line (map remove-comments lines)
-          :when (not (empty? line))]
-      (let [{:keys [namespace service-name]} (parse-bootstrap-line! line)]
-        (resolve-service! namespace service-name)))))
+  service graph that is the result of merging the graphs of all of the
+  services specified in the configuration files."
+  ; We remove the duplicate entries to allow the user to have duplicate entries in their
+  ; bootstrap files. If we didn't remove them, it would look like two services were trying
+  ; to implement the same protocol when we check for duplicate service implementations
```
rlinehan: I'm really confused by this comment - don't we not want to allow users to have duplicate entries in their bootstrap files?

dev-resources/bootstrapping/cli/bootstrap_with_comments.cfg
```diff
@@ -0,0 +1,5 @@
+# commented out line
+puppetlabs.trapperkeeper.examples.bootstrapping.test-services/hello-world-service # comment
+; another commented out line
+ ;puppetlabs.trapperkeeper.examples.bootstrapping.test-services/foo-test-service
```
rlinehan: This has a space before the comment - not sure that it matters? but figured I would point it out cuz it seems a little inconsistent.

test/puppetlabs/trapperkeeper/bootstrap_test.clj
```diff
@@ -262,25 +244,152 @@ puppetlabs.trapperkeeper.examples.bootstrapping.test-services/foo-test-service ;
           hello-world-svc (get-service app :HelloWorldService)]
       (is (= (hello-world hello-world-svc) "hello world")))))
 
+(deftest duplicate-service-definitions
+  (testing "Duplicate service definitions causes error with filename and line numbers"
```
rlinehan: Would it make sense to have a test of what happens if there are duplicate service definitions between two different files? Not sure that with the code paths that that would be testing anything different, but it is a scenario that I find myself wondering about as I review this PR.

src/puppetlabs/trapperkeeper/bootstrap.clj
jpinsonault: I can't find a jira paper trail for this, I think @cprice404 and I just talked about it irl, but we were thinking that TK should be accommodating if the same service entry appeared in two different bootstraps, or twice in one entry. A test to verify that was added in #231. I'm definitely open to being more strict though if that's what people think is better, I don't have a strong opinion either way

camlow325: I think the rationale for this is that we'll probably have cross-over points where we want to move a bootstrap service entry from one file to another one - like with Puppet Server where we'll want to move the `certificate-authority` service entry from the main `bootstrap.cfg` file over to the new CA-specific `bootstrap.cfg` one (or whatever we end up calling it) - and we didn't necessarily want to cause the app to fail to launch just because we happened to have references to exactly the same service namespace in two different files - like one for the new file and one from the old file.  This would also seem to support the use cases being discussed in SERVER-1213 where some users might want to prepare for an upgrade early by laying down file content as they eventually want to have it after the upgrade living alongside the content as it needs to exist in order to work with the old release still being used - i.e., the monolithic bootstrap.cfg that you have to have today.  @cprice404 may want to provide more color on this when he gets back.  In any event, this behavior would at least be backward compatible with the current Trapperkeeper implementation.

rlinehan: Ah, I see. I was reading "duplicate entries" as meaning "duplicate protocol implementations" (I think because that's what "duplicate <some other words I wasn't paying full attention to>" means everywhere else). This makes much more sense. Maybe we could use a term other than "duplicate entries" to make this a little clear? IDK what though. Feel free to ignore this because obviously I can't read.

dev-resources/bootstrapping/cli/bootstrap_with_comments.cfg
jpinsonault: I just copied this from a test, but it seemed intentional, to prove that it could handle lines with spaces and comments I guess?

test/puppetlabs/trapperkeeper/bootstrap_test.clj
jpinsonault: It shouldn't be treated differently/go down different code paths, but it certainly couldn't hurt, it might not always be that way

src/puppetlabs/trapperkeeper/bootstrap.clj
jpinsonault: It confuses me too, I'd be happy to change the name but I can't think of a better one off the top of my head either

camlow325:
:+1: I think @rlinehan's comments are a superset of mine at this point so I'm happy merging this whenever her concerns have been addressed.

dev-resources/bootstrapping/cli/bootstrap_with_comments.cfg
rlinehan: Ah cool, that makes sense.

src/puppetlabs/trapperkeeper/bootstrap.clj
rlinehan: Hmm... how about something like

```
We remove the duplicate entries (same exact name) to allow the user to have duplicate entries
in their bootstrap files. If we didn't remove them, it would look like two services were trying
to implement the same protocol when we check for duplicate service implementations, and we
 want to allow entries with the same exact name in order to support workflows where users are
preparing to upgrade and might copy an entry to another file.
```

camlow325: FWIW, 'duplicate entries' seems like a reasonable way to refer to this for name.  All of the other things I've thought of are much wordier and probably not any more helpful.

rlinehan: Cool, I'm fine with keeping it as-is if y'all are - I leave it up to your judgement @jpinsonault.

jpinsonault:
@rlinehan @camlow325 I think I've addressed all your comments

rlinehan:
👍 LGTM!

PR #234, opened 2016-04-12
(TK-362) Honor log level in logged? macro

justinstoller:
Previously we ignored the log-level parameter if it was supplied.

This adds test cases to ensure that logged? fails, as well as succeeds,
properly and updates the logged? matching mechanism to first filter logs
by log level if requested.


fpringvaldsen:
:+1:

PR #235, opened 2016-04-14
(TK-349) Log warning when service can't be found

jpinsonault:
This commit changes the behavior of TK when a bootstrap file contains a service
entry that can't be resolved. Previous TK would throw an error and fail to
start. Now TK will log a warning and attempt to continue loading the app

This is being added to support workflows where a user might want to add a
service entry to their bootstrap before a TK app actually requires it.

TK continues to throw errors and abort if the service entry is malformed or if
the service is found, but doesn't have a valid service graph

### DIFF
--- a/src/puppetlabs/trapperkeeper/bootstrap.clj
+++ b/src/puppetlabs/trapperkeeper/bootstrap.clj
@@ -45,23 +45,6 @@
                            line
                            "\n\nAll lines must be of the form: '<namespace>/<service-fn-name>'.")})))
 
-(schema/defn ^:private resolve-service! :- (schema/protocol services/ServiceDefinition)
-  "Given the namespace and name of a service, loads the namespace,
-  calls the function, validates that the result is a valid service definition, and
-  returns the service definition.  Throws an `IllegalArgumentException` if the
-  service definition cannot be resolved."
-  [resolve-ns :- schema/Str
-   service-name :- schema/Str]
-  (try (require (symbol resolve-ns))
-       (catch FileNotFoundException e
-         (throw+ {:type ::bootstrap-parse-error
-                  :message (str "Unable to load service: " resolve-ns "/" service-name)
-                  :cause e})))
-  (if-let [service-def (ns-resolve (symbol resolve-ns) (symbol service-name))]
-    (internal/validate-service-graph! (var-get service-def))
-    (throw+ {:type ::bootstrap-parse-error
-             :message (str "Unable to load service: " resolve-ns "/" service-name)})))
-
 (schema/defn ^:private remove-comments :- schema/Str
   "Given a line of text from the bootstrap config file, remove
   anything that is commented out with either a '#' or ';'. If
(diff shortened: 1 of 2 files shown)


jpinsonault:
This is based on #233, so it shouldn't be merged until that PR is. Only the last commit, `c25e7ee` is relevant to this PR

src/puppetlabs/trapperkeeper/bootstrap.clj
```diff
@@ -221,6 +204,23 @@
     (if-let [duplicate (first (find-duplicates services services/service-def-id))]
       (throw (duplicate-protocol-error (key duplicate) (val duplicate) service->entry-map)))))
 
+(schema/defn ^:private resolve-service! :- (schema/protocol services/ServiceDefinition)
+  "Given the namespace and name of a service, loads the namespace,
+  calls the function, validates that the result is a valid service definition, and
+  returns the service definition.  Throws an `IllegalArgumentException` if the
```
fpringvaldsen: This is no longer throwing an `IllegalArgumentException`

fpringvaldsen:
:+1: I just had one minor comment

rlinehan:
I'm 👍 on this once #233 has been merged (and probably this will need to be rebased?).

src/puppetlabs/trapperkeeper/bootstrap.clj
```diff
@@ -230,24 +230,36 @@
...
+  If the bootstrap entry can't be resolved, logs a warning and returns nil.
+
+  Throws an IllegalArgumentException if there is a problem parsing the bootstrap
+  entry, or if the service is found but it has an invalid service graph."
+  [{:keys [bootstrap-file line-number entry] :as bootstrap-entry} :- AnnotatedBootstrapEntry]
+  (try+
+    (let [{:keys [namespace service-name]} (parse-bootstrap-line! entry)]
+      (resolve-service! namespace service-name))
+    (catch [:type ::missing-service] {:keys [message]}
+      (log/warnf "%s:%s\nUnable to load service '%s'" bootstrap-file line-number entry))
```
camlow325: With the current code, I get error messages like this when there's a problem, which I find a little hard to read:

```
2016-04-18 15:55:34,576 WARN  [p.t.bootstrap] ./examples/java_service/bootstrap.cfg:2
Unable to load service 'crappy/service'
```

I find it a little more readable to have the error message lead off with "Unable to load service ... " and have the service name, bootstrap file, and line number follow.  Having it all one line would seem reasonable to me as well.

src/puppetlabs/trapperkeeper/bootstrap.clj
```diff
@@ -230,24 +230,36 @@
...
    original-message :- schema/Str]
   (IllegalArgumentException.
-   (format (str "Problem loading service '%s' on line '%s' in bootstrap "
-                "configuration file '%s':\n%s")
-           entry line-number bootstrap-file original-message)))
+   (format (str "%s:%s\nProblem loading service '%s':\n%s")
+           bootstrap-file line-number entry original-message)))
+
+(schema/defn resolve-and-handle-errors! :- (schema/maybe (schema/protocol services/ServiceDefinition))
+  "Attemps to resolve a bootstrap entry into a ServiceDefinition.
```
camlow325: s/Attemps/Attempts

camlow325:
:+1: from me also once #233 is merged and this is rebased on top of it.  Would like to hear @rlinehan's feedback on the format of the log messages.  I don't have a strong opinion on those so if ye are good with them as they are I am too.

src/puppetlabs/trapperkeeper/bootstrap.clj
rlinehan: Ah, I didn't catch this when I was reviewing, but I agree that I think having `Unable to load service` and everything on one line seems better. Maybe `Unable to load service 'crappy/service' from <file>:<line>`?

camlow325: @rlinehan, that format sounds good to me, thanks.  @jpinsonault, if you agree with that, I'd suggest applying that here as well as for the "Problem loading service" exception which is thrown from `bootstrap-error` for problems with basic bootstrap parsing and creation of the service graph.

jpinsonault: Agreed, sounds good to me

jpinsonault:
Rebased and addressed feedback

rlinehan:
I'm 👍 but looks like this needs to be rebased again now that #233 has been merged. Sorry @jpinsonault!

jpinsonault:
Rebased

rlinehan:
:shipit:

PR #241, opened 2016-05-17
(TK-375) Fix bootstrap regression

jpinsonault:
TK 1.4.0 introduced a regression which caused TK to fail to start if a
bootstrap file was found in a jar file on the class path.

This commit restores that functionality and adds a couple tests for the feature


cprice404:
+1

camlow325:
👍 LGTM

PR #243, opened 2016-06-29
(TK-309) Schemafy TK service macro

janelu2:
This purpose of this commit is to:
- remove :pre and :post conditions 
- collapse validate-fn-forms and validate-deps-form 
- use schema to verify input/ouput

### DIFF
--- a/src/puppetlabs/trapperkeeper/services_internal.clj
+++ b/src/puppetlabs/trapperkeeper/services_internal.clj
@@ -58,6 +58,13 @@
    :dependencies (schema/pred sequential?)
    :fns-map FnsMap})
 
+(def ProtocolandDependenciesMap
+  "Schema defining the map used to represent the protocol and dependencies
+  of a service."
+  {:fns (schema/pred seq?)
+   :dependencies (schema/pred vector?)
+   :service-protocol-sym (schema/maybe Symbol)})
+
 (def ServiceMap
   "Schema defining the map used to represent a schema in the output service
    map."
@@ -87,14 +94,13 @@
             optional-form (array-map dep optional-dep-default)]
         (recur (rest optional) (conj output optional-form))))))
 
-(schema/defn ^:always-validate find-prot-and-deps-forms!
-  :- {:fns (schema/pred seq?) :dependencies (schema/pred vector?) :service-protocol-sym (schema/maybe Symbol)}
+(schema/defn ^:always-validate find-prot-and-deps-forms! :- ProtocolandDependenciesMap
   "Given the forms passed to the service macro, find the service protocol
   (if one is provided), the dependency list, and the function definitions.
   Throws `IllegalArgumentException` if the forms do not represent a valid service.
   Returns a map containing the protocol, dependency list, and fn forms."
   [forms :- (schema/pred seq?)]
-  (let [f (if (some true? ((juxt symbol? map? vector?) (first forms)))
+  (let [f (if ((some-fn symbol? map? vector?) (first forms))
             (first forms)
             (throw (IllegalArgumentException.
                     (format
--- a/src/puppetlabs/trapperkeeper/services_internal.clj
+++ b/src/puppetlabs/trapperkeeper/services_internal.clj
@@ -61,7 +61,7 @@
 (def ProtocolandDependenciesMap
   "Schema defining the map used to represent the protocol and dependencies
   of a service."
-  {:fns (schema/pred seq?)
+  {:fns [(schema/pred seq?)]
    :dependencies (schema/pred vector?)
    :service-protocol-sym (schema/maybe Symbol)})
 
--- a/src/puppetlabs/trapperkeeper/services_internal.clj
+++ b/src/puppetlabs/trapperkeeper/services_internal.clj
@@ -87,64 +87,36 @@
             optional-form (array-map dep optional-dep-default)]
         (recur (rest optional) (conj output optional-form))))))
 
-;; TODO Not converting to use schema since this fn's behavior should literally
-;; be a schema
-(defn validate-fn-forms!
-  "Validate that all of the fn forms in the service body appear to be
-  valid fn definitions.  Throws `IllegalArgumentException` otherwise."
-  [fns]
-  {:pre [(seq? fns)]
-   :post [(map? %)
-          (= #{:fns} (set (keys %)))]}
-  (if (every? seq? fns)
-    {:fns fns}
-    (throw (IllegalArgumentException.
-            (format
-             "Invalid service definition; expected function definitions following dependency list, invalid value: '%s'"
-             (pr-str (first (filter #(not (seq? %)) fns))))))))
-
-;; TODO Not converting to use schema since this fn's behavior should literally
-;; be a schema
-(defn validate-deps-form!
-  "Validate that the service body has a valid dependency specification.
-  Throws `IllegalArgumentException` otherwise."
-  [forms]
-  {:pre [(seq? forms)]
-   :post [(map? %)
-          (= #{:fns :dependencies} (set (keys %)))]}
-  (let [f (first forms)
-        deps (cond
-               (vector? f) f
-               (map? f) (transform-deps-map f)
-               :else (throw (IllegalArgumentException.
-                             (format
-                              "Invalid service definition; expected dependency list following protocol, found: '%s'"
-                              (pr-str f)))))]
-    (merge {:dependencies deps} (validate-fn-forms! (rest forms)))))
-
-;; TODO Not converting to use schema since this fn's behavior should literally
-;; be a schema
-(defn find-prot-and-deps-forms!
+(schema/defn ^:always-validate find-prot-and-deps-forms!
+  :- {:fns (schema/pred seq?) :dependencies (schema/pred vector?) :service-protocol-sym (schema/maybe Symbol)}
   "Given the forms passed to the service macro, find the service protocol
   (if one is provided), the dependency list, and the function definitions.
   Throws `IllegalArgumentException` if the forms do not represent a valid service.
   Returns a map containing the protocol, dependency list, and fn forms."
-  [forms]
-  {:pre [(seq? forms)]
-   :post [(map? %)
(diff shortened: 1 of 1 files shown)


janelu2:
@cprice404 @nathanielksmith please take a look whenever you guys have time :smile: let me know anything that I can do to improve what is currently there.

janelu2:
just added back in the schema comment, somehow lost that in my original commit

src/puppetlabs/trapperkeeper/services_internal.clj
```diff
@@ -87,64 +87,36 @@
...
-                             (format
-                              "Invalid service definition; expected dependency list following protocol, found: '%s'"
-                              (pr-str f)))))]
-    (merge {:dependencies deps} (validate-fn-forms! (rest forms)))))
-
-;; TODO Not converting to use schema since this fn's behavior should literally
-;; be a schema
-(defn find-prot-and-deps-forms!
+(schema/defn ^:always-validate find-prot-and-deps-forms!
+  :- {:fns (schema/pred seq?) :dependencies (schema/pred vector?) :service-protocol-sym (schema/maybe Symbol)}
```
cprice404: It's probably worth extracting this schema to be a named schema at the top of the file and then referring to the name here.  Maybe call it ProtocolAndDependencies or ProtocolAndDependenciesForms or something.

src/puppetlabs/trapperkeeper/services_internal.clj
```diff
@@ -87,64 +87,36 @@
...
-          (vector? (:dependencies %))
-          (seq? (:fns %))]}
-  (let [f (first forms)]
-    (cond
-      (symbol? f) (merge {:service-protocol-sym f} (validate-deps-form! (rest forms)))
-      (map? f) (merge {:service-protocol-sym nil} (validate-deps-form! forms))
-      (vector? f) (merge {:service-protocol-sym nil} (validate-deps-form! forms))
-      :else (throw (IllegalArgumentException.
+  [forms :- (schema/pred seq?)]
+  (let [f (if (some true? ((juxt symbol? map? vector?) (first forms)))
```
cprice404: I think on this line you could do:

    (if ((some-fn symbol? map? vector?) (first forms))

cprice404:
This looks pretty good to me other than minor nitpicks.

I'm curious what we defined as the scope of this ticket/work, though.  I know that those two fns are the most egregious places where we were using pre/post-conditions, but there are some other ones in `core` and `internal` that could possibly also be done.  I'm fine either way, not sure if I know what all @nathanielksmith had been thinking we'd tackle in this round.

janelu2:
@cprice404 great, I've made those changes. I've been thinking about changing `f` and `ff` to something more descriptive so that it is less confusing but I can't really think of anything great. Any ideas?

cprice404:
Cool, I think I'm +1 on this.  Let's see if @nathanielksmith is back to do a second review tomorrow.

src/puppetlabs/trapperkeeper/services_internal.clj
```diff
@@ -58,6 +58,13 @@
    :dependencies (schema/pred sequential?)
    :fns-map FnsMap})
 
+(def ProtocolandDependenciesMap
+  "Schema defining the map used to represent the protocol and dependencies
+  of a service."
+  {:fns (schema/pred seq?)
```
cprice404: Technically, this could be a little more strict.  We know that, at least, they should be a seq of seqs, so you could maybe do `(schema/pred #(and (seq? %) (every? seq? %)))`.  Since `every?` would probably fail on anything that isn't a seq/coll, it might be ok to just do `(schema/pred #(every? seq? %))`.

You could try to get even more advanced inside of there, since we kind of know more about the specific forms that are necessary for a seq to represent a fn def, like `(symbol? (first %))`, and `(vec? (second %))` (i.e., to match a fn definition like `(foo [arg1 arg2] ...)`, but I'm not sure that's necessary (we weren't doing it before).

src/puppetlabs/trapperkeeper/services_internal.clj
```diff
@@ -58,6 +58,13 @@
...
+   :dependencies (schema/pred vector?)
```
cprice404: This could also probably be made to be more strict, since we know exactly what syntaxes are legal binding forms for the dependencies.  Binding forms can look like this:

```clj
[[:SomeService some-fn other-fn]
 OtherService]
```

So you could do something like `[BindingForm]`, and then define the `BindingForm` schema as something that used schema/cond or schema/or to represent the two legal forms, which would probably be just `schema/Symbol` for the second case, and then something along the lines of `[(schema/one schema/Keyword) schema/Symbol]` for the first case.

cprice404:
I left a couple of notes on how that output schema could be made to be more strict.  It seems like maybe the part about ensuring that `:fns` is a seq of seqs might be worthwhile; the others I don't have a strong opinion about.

tbh if I was going to invest more time in this I'd *probably* lean toward looking into some of the other pre/post conditions in files like `core.clj` and `internal.clj` over tightening up the output schema here, but I don't really have a strong opinion on that either.

src/puppetlabs/trapperkeeper/services_internal.clj
nathanielksmith: I think you can use `[clojure.lang.Cons]` for `:fns`.

src/puppetlabs/trapperkeeper/services_internal.clj
nathanielksmith: We already have a schema for this: `DependencyMap`. Should be able to use that, here?

src/puppetlabs/trapperkeeper/services_internal.clj
```diff
@@ -87,64 +94,35 @@
...
+        service-protocol-sym (if (symbol? f) f nil)
+        forms (if (nil? service-protocol-sym) forms (rest forms))
+        ff (first forms)
+        deps (cond (vector? ff) ff
+                   (map? ff) (transform-deps-map ff)
+                   :else (throw (IllegalArgumentException.
+                                 (format
+                                  "Invalid service definition; expected dependency list following protocol, found: '%s'"
+                                  (pr-str ff)))))]
+    (if (every? seq? (rest forms))
```
nathanielksmith: Given a tighter output schema, we'll now get a schema failure if (rest forms) isn't what we want. On the up side, we can avoid this check and the resulting exception; on the down side, we are trading a pretty specific error for a schema error. I feel like using schema comes with this kind of tradeoff and I'm okay with it; but @cprice404 ?

cprice404: I would greatly prefer to keep the more user-friendly exception in place.

src/puppetlabs/trapperkeeper/services_internal.clj
nathanielksmith: oh i fucked up and read some schemas backwards, nevermind

nathanielksmith: I donno if it's worth it to check that we're creating the right binding forms for fnk; i'm happy keeping the status quo of just checking vector (with like [schema/Any] or something)

nathanielksmith:
I'm :+1: on this and don't have any more changes to suggest, would be happy with a merge.

janelu2:
Did some fixup based on some of the comments given, too tired to really dive into that [BindingForms] rabbithole today. I think it's good, lmk if there's anything else!

src/puppetlabs/trapperkeeper/services_internal.clj
```diff
@@ -58,6 +58,13 @@
    :dependencies (schema/pred sequential?)
    :fns-map FnsMap})
 
+(def ProtocolandDependenciesMap
+  "Schema defining the map used to represent the protocol and dependencies
+  of a service."
+  {:fns [clojure.lang.PersistentList]
```
cprice404: this is probably ok, but it seems like it might be overly specific and it'd be nicer if we didn't have to reference a Java class.  I'm guessing that you tried some other things like `(schema/pred seq?)` and they didn't work?

cprice404: @nathanielksmith I was chatting with @janelu2 and she said that you had a preference for the Java class here.  I'm not necessarily opposed, but am interested in hearing your reasoning.

cprice404: @nathanielksmith any thoughts on this?

nathanielksmith: it's a personal preference thing, my opinions are not Strong.

cprice404: ok, but is there something about using the Java class here that is compelling to you?  Like... does it catch/enforce something that we'd miss if we did something more clojure-y?  I just want to make sure I'm not overlooking something that you thought of.

cprice404: @nathanielksmith ping.  Would like to get this merged; I have a slight pref for changing it to something more clojure-y but would be OK with this version if there's something you thought of that would warrant it.

haus: fwiw, i tend to agree with @cprice404 here.

janelu2:
@cprice404 final peek?

src/puppetlabs/trapperkeeper/services_internal.clj
```diff
@@ -61,7 +61,7 @@
 (def ProtocolandDependenciesMap
   "Schema defining the map used to represent the protocol and dependencies
   of a service."
-  {:fns [clojure.lang.PersistentList]
+  {:fns (schema/pred seq?)
```
cprice404: So this just checks that `fns` is a sequence, rather than a sequence of sequences.  I think you want it to be `[(schema/pred seq?)]`?

janelu2: shit that got lost in translation, fixed

cprice404:
lgtm, will merge after travis

PR #247, opened 2016-07-19
(TK-382) Fix defservice + optional deps

nathanielksmith:
This commit fixes defservice's sadness when specifying optional
dependencies without a protocol. The bug was from use of
clojure.tools.macro/name-with-attributes, which was mistaking the
dependency map for metadata and removing it from the forms passed on to
the service macro itself.

The remediation was to implement our own name-with-attributes that is
smart enough to ignore dependency maps and not blindly accept them as
metadata maps.

This commit also adds tests for the various permutations of input to
defservice.

### DIFF
--- a/test/puppetlabs/trapperkeeper/optional_deps_test.clj
+++ b/test/puppetlabs/trapperkeeper/optional_deps_test.clj
@@ -32,6 +32,100 @@
                                             couplet))))
 
 (deftest optional-deps-test
+  (testing "when using defservice"
+    (testing "with destructuring"
+      (testing "and protocol"
+        (defservice poetry-svc-w-destructure-and-proto
+          PoetryService
+          {:required [[:HaikuService haiku]]
+           :optional [SonnetService]}
+          (init [this ctx] ctx)
+          (get-haiku [this] (haiku "meh"))
+          (get-sonnet [this] "feh"))
+        (is (build-app [poetry-svc-w-destructure-and-proto haiku-service] {})))
+      (testing "and no protocol"
+        (defservice poetry-svc-w-destructure
+          {:required [[:HaikuService haiku]]
+           :optional [SonnetService]}
+          (init [this ctx] ctx))
+        (is (build-app [poetry-svc-w-destructure haiku-service] {}))))
+
+    (testing "with a protocol"
+      (defservice poetry-svc-w-protocol
+        PoetryService
+        {:required [HaikuService]
+         :optional [SonnetService]}
+        (init [this ctx] ctx)
+        (get-haiku [this] "haiku")
+        (get-sonnet [this] "sonnet"))
+      (is (build-app [poetry-svc-w-protocol haiku-service] {})))
+
+    (testing "without a protocol"
+      (defservice poetry-svc-wo-protocol
+        {:required [HaikuService]
+         :optional [SonnetService]}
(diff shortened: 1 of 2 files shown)


test/puppetlabs/trapperkeeper/optional_deps_test.clj
```diff
@@ -32,6 +32,83 @@
...
 (deftest optional-deps-test
+  (testing "when using defservice"
+    (testing "with a protocol"
+      (defservice poetry-svc-w-protocol
+        PoetryService
+        {:required [HaikuService]
+         :optional [SonnetService]}
+        (init [this ctx] ctx)
+        (get-haiku [this] "haiku")
+        (get-sonnet [this] "sonnet"))
```
cprice404: This is a nitpick and a stylistic thing that I won't mind if I'm overruled on, but, I think it's weird to call `def*` things inside of other forms.  So I'd probably write these tests such that the `defservice` was at the top level of the file, just prior to the test that was consuming it.

nathanielksmith: I also hate this. But the point of these tests is to make sure that defservice compiles ok. I figured having them fail during macro compilation near an actual (testing) line would make the horrific task of debugging test failures here slightly less bad.

cprice404:
Might be good to add another test or two that cover the destructured syntax `[[:SonnetService sonnet]]`.  I don't think it needs to be as exhaustive as what you have in this PR but just one or two cases.

cprice404:
+1 other than suggestion for additional tests and possibly moving the `defservice` forms up to the root.

nathanielksmith:
@cprice404 there are already such tests in this namespace. they don't use defservice because at that point parsing the dependency map happens in the service macro.

test/puppetlabs/trapperkeeper/optional_deps_test.clj
cprice404: :sake: i'm ok with it as-is if i'm the only one who'd do it the other way.

cprice404:
@nathanielksmith but that's an implementation detail, right?  It seems like, since this bug was around a difference between `defservice` and `service` behavior, it couldn't hurt to have at least one test that had a defservice with a destructuring?

test/puppetlabs/trapperkeeper/optional_deps_test.clj
nathanielksmith: @haus thoughts?

nathanielksmith:
touche; i'll add one.

test/puppetlabs/trapperkeeper/optional_deps_test.clj
haus: I've generally tried to avoid putting defs in tests, but I don't feel strongly about it. If it seems likely we'll have other tests also using the same service it's probably worth lifting it out.

nathanielksmith: these services are just for testing macro compilation and nothing else.

haus: Having them in the test that's using them seems fine to me then.

cprice404:
+1

haus:
😢

PR #248, opened 2016-07-25
(TK-345) add support for optional restart-file

janelu2:
This PR adds support for using a restart file that holds nothing but a long as a counter. This counter will increment each time a service completes its starting process.

### DIFF
--- a/src/puppetlabs/trapperkeeper/internal.clj
+++ b/src/puppetlabs/trapperkeeper/internal.clj
@@ -49,21 +49,23 @@
 
 (defn inc-restart-counter!
   "Increments the counter in a restart-file for purposes of supporting HUP behavior"
-  [restart-file-path]
-  (try
-    (if (fs/exists? restart-file-path)
-      (if (and (fs/writeable? restart-file-path) (fs/readable? restart-file-path))
-        (spit restart-file-path (inc (Long/parseLong (slurp restart-file-path))))
-        (throw (IllegalStateException. (format "Restart file %s is not readable and/or writeable" restart-file-path))))
-      (let [dir-path (fs/parent restart-file-path)]
-        (fs/mkdirs dir-path)
-        (spit restart-file-path "1")))
-    (catch ArithmeticException e
-      (spit restart-file-path "1")
-      (log/debug "Number of restarts has exceeded Long/MAX_VALUE, resetting file to 1"))
-    (catch NumberFormatException e
-      (spit restart-file-path "1")
-      (log/error "Restart file is unparseable, resetting file to 1"))))
+  [app]
+  (when-let
+      [restart-file-path (get-in-config (a/get-service app :ConfigService) [:global :restart-file])]
+      (try
+        (if (fs/exists? restart-file-path)
+          (if (and (fs/writeable? restart-file-path) (fs/readable? restart-file-path))
+            (spit restart-file-path (inc (Long/parseLong (slurp restart-file-path))))
+            (throw (IllegalStateException. (format "Restart file %s is not readable and/or writeable" restart-file-path))))
+          (let [dir-path (fs/parent restart-file-path)]
+            (fs/mkdirs dir-path)
+            (spit restart-file-path "1")))
+        (catch ArithmeticException e
+          (spit restart-file-path "1")
+          (log/debug "Number of restarts has exceeded Long/MAX_VALUE, resetting file to 1"))
+        (catch NumberFormatException e
+          (spit restart-file-path "1")
+          (log/error "Restart file is unparseable, resetting file to 1")))))
 
 (defn validate-service-graph!
   "Validates that a ServiceDefinition contains a valid trapperkeeper service graph.
@@ -583,8 +585,7 @@
         this)
       (a/start [this]
         (run-lifecycle-fns app-context s/start "start" ordered-services)
-        (when-let [restart-file (get-in-config (a/get-service this :ConfigService) [:global :restart-file])]
-          (inc-restart-counter! restart-file))
+        (inc-restart-counter! this)
         this)
       (a/stop [this]
         (shutdown! app-context)
@@ -595,8 +596,7 @@
           (doseq [svc-id (keys services-by-id)] (swap! app-context assoc-in [:service-contexts svc-id] {}))
           (run-lifecycle-fns app-context s/init "init" ordered-services)
           (run-lifecycle-fns app-context s/start "start" ordered-services)
-          (when-let [restart-file (get-in-config (a/get-service this :ConfigService) [:global :restart-file])]
-            (inc-restart-counter! restart-file))
+          (inc-restart-counter! this)
           this
           (catch Throwable t
             (deliver shutdown-reason-promise {:cause :service-error
--- a/src/puppetlabs/trapperkeeper/internal.clj
+++ b/src/puppetlabs/trapperkeeper/internal.clj
@@ -1,11 +1,13 @@
 (ns puppetlabs.trapperkeeper.internal
-  (:import (clojure.lang ExceptionInfo IFn IDeref))
+  (:import (clojure.lang ExceptionInfo IFn IDeref)
+           (java.lang ArithmeticException NumberFormatException)
+           (java.io FileNotFoundException))
   (:require [clojure.tools.logging :as log]
             [beckon]
             [plumbing.graph :as graph]
             [slingshot.slingshot :refer [throw+]]
             [puppetlabs.kitchensink.core :refer [add-shutdown-hook! boolean? cli!]]
-            [puppetlabs.trapperkeeper.config :refer [config-service]]
+            [puppetlabs.trapperkeeper.config :refer [config-service get-in-config]]
             [puppetlabs.trapperkeeper.app :as a]
             [puppetlabs.trapperkeeper.common :as common]
             [puppetlabs.trapperkeeper.services :as s]
@@ -13,7 +15,8 @@
             [schema.core :as schema]
             [clojure.core.async :as async]
             [clojure.core.async.impl.protocols :as async-prot]
-            [clojure.string :as string]))
+            [clojure.string :as string]
+            [me.raynes.fs :as fs]))
 
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;; Schemas
@@ -43,6 +46,24 @@
    (map? service-graph)
    (every? keyword? (keys service-graph))
    (every? (some-fn ifn? service-graph?) (vals service-graph))))
+
+(defn inc-restart-counter!
+  "Increments the counter in a restart-file for purposes of supporting HUP behavior"
+  [restart-file-path]
+  (try
+    (if (fs/exists? restart-file-path)
+      (if (and (fs/writeable? restart-file-path) (fs/readable? restart-file-path))
+        (spit restart-file-path (inc (Long/parseLong (slurp restart-file-path))))
+        (throw (IllegalStateException. (format "Restart file %s is not readable and/or writeable" restart-file-path))))
+      (let [dir-path (fs/parent restart-file-path)]
+        (fs/mkdirs dir-path)
+        (spit restart-file-path "1")))
+    (catch ArithmeticException e
+      (spit restart-file-path "1")
+      (log/debug "Number of restarts has exceeded Long/MAX_VALUE, resetting file to 1"))
+    (catch NumberFormatException e
+      (spit restart-file-path "1")
+      (log/error "Restart file is unparseable, resetting file to 1"))))
 
 (defn validate-service-graph!
   "Validates that a ServiceDefinition contains a valid trapperkeeper service graph.
@@ -562,6 +583,8 @@
         this)
       (a/start [this]
         (run-lifecycle-fns app-context s/start "start" ordered-services)
+        (when-let [restart-file (get-in-config (a/get-service this :ConfigService) [:global :restart-file])]
+          (inc-restart-counter! restart-file))
         this)
       (a/stop [this]
         (shutdown! app-context)
@@ -572,6 +595,8 @@
           (doseq [svc-id (keys services-by-id)] (swap! app-context assoc-in [:service-contexts svc-id] {}))
           (run-lifecycle-fns app-context s/init "init" ordered-services)
           (run-lifecycle-fns app-context s/start "start" ordered-services)
+          (when-let [restart-file (get-in-config (a/get-service this :ConfigService) [:global :restart-file])]
+            (inc-restart-counter! restart-file))
           this
           (catch Throwable t
             (deliver shutdown-reason-promise {:cause :service-error
--- a/test/puppetlabs/trapperkeeper/services_test.clj
+++ b/test/puppetlabs/trapperkeeper/services_test.clj
@@ -3,15 +3,17 @@
             [puppetlabs.trapperkeeper.services :refer
              [defservice service] :as svcs]
             [puppetlabs.trapperkeeper.app :as app]
+            [puppetlabs.kitchensink.core :as kitchensink]
             [puppetlabs.trapperkeeper.testutils.bootstrap :refer
              [bootstrap-services-with-empty-config
-              with-app-with-empty-config]]
+              with-app-with-empty-config
+              with-app-with-config]]
             [schema.test :as schema-test]
             [puppetlabs.kitchensink.testutils.fixtures :refer [with-no-jvm-shutdown-hooks]]
             [puppetlabs.trapperkeeper.internal :as internal]
-            [puppetlabs.trapperkeeper.testutils.logging :refer [with-test-logging]]
             [puppetlabs.trapperkeeper.core :as tk]
-            [puppetlabs.kitchensink.testutils :as ks-testutils])
+            [puppetlabs.kitchensink.testutils :as ks-testutils]
+            [me.raynes.fs :as fs])
   (:import (java.util.concurrent ExecutionException)))
 
 (use-fixtures :once schema-test/validate-schemas with-no-jvm-shutdown-hooks)
@@ -61,6 +63,67 @@
                      (service1-fn [this] "hi"))
           app (bootstrap-services-with-empty-config [service1])]
       (is (not (nil? app))))))
+
+(deftest test-restart-file-correct-count
+  (testing "create a restart file and check that it correctly increments when a service starts"
+    (let [service1 (service Service1
+                            []
+                            (service1-fn [this] "hi"))
+          temp-file (kitchensink/temp-file-name "counter")]
+          (with-app-with-config app [service1] {:global {:restart-file temp-file}}
+            (app/restart app)
+            (is (= (slurp temp-file) "2"))))))
+
+(deftest test-restart-file-HUP-restart
+  (testing "check that restart file correctly increments on HUP restarts"
+    (let [service1 (service Service1
+                            []
+                            (service1-fn [this] "hi"))
+          temp-file (kitchensink/temp-file-name "counter")]
+      (with-app-with-config app [service1] {:global {:restart-file temp-file}}
+        (app/restart app)
+        (is (= (slurp temp-file) "2"))))))
+
+(deftest test-restart-file-big-int-limit
+  (testing "restart file should reset to 1 if the counter is too large"
+    (let [service1 (service Service1
+                            []
+                            (service1-fn [this] "hi"))
+          temp-file (kitchensink/temp-file-name "counter")]
+      (spit temp-file "9223372036854775807")
+      (with-app-with-config app [service1] {:global {:restart-file temp-file}}
+        (is (= (slurp temp-file) "1")))
+      (with-app-with-config app [service1] {:global {:restart-file temp-file}}
+        (is (= (slurp temp-file) "2"))))))
+
+(deftest test-invalid-restart-file
+  (testing "restart file should reset to 1 if the file is unparesable"
+    (let [service1 (service Service1
+                            []
+                            (service1-fn [this] "hi"))
+          temp-file (kitchensink/temp-file-name "counter")]
+      (spit temp-file "hello")
+      (with-app-with-config app [service1] {:global {:restart-file temp-file}}
+        (is (= (slurp temp-file) "1"))))))
+
+(deftest test-restart-file-no-permissions
+  (testing "exception should be thrown if restart file is not readable/writeable"
+    (let [service1 (service Service1
+                            []
+                            (service1-fn [this] "hi"))
+          temp-file (kitchensink/temp-file-name "counter")]
+      (fs/chmod "u-rw" (fs/touch temp-file))
+      (is (thrown? IllegalStateException
+                   (with-app-with-config app [service1] {:global {:restart-file temp-file}}))))))
+
+(deftest test-restart-file-missing-parent-dirs
+  (testing "some portion of the parent directory structure does not exist"
+    (let [service1 (service Service1
+                            []
+                            (service1-fn [this] "hi"))
+          temp-file (fs/file (kitchensink/temp-dir "foo") "bar" "baz" "counter")]
+      (with-app-with-config app [service1] {:global {:restart-file temp-file}}
+        (is (= (slurp temp-file) "1"))))))
 
 (defn create-lifecycle-services
   [call-seq]
--- a/test/puppetlabs/trapperkeeper/services_test.clj
+++ b/test/puppetlabs/trapperkeeper/services_test.clj
@@ -71,8 +71,7 @@
                             (service1-fn [this] "hi"))
           temp-file (kitchensink/temp-file-name "counter")]
           (with-app-with-config app [service1] {:global {:restart-file temp-file}}
-            (app/restart app)
-            (is (= (slurp temp-file) "2"))))))
+            (is (= (slurp temp-file) "1"))))))
 
 (deftest test-restart-file-HUP-restart
   (testing "check that restart file correctly increments on HUP restarts"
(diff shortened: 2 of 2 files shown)


janelu2:
@cprice404 let me know your thoughts

src/puppetlabs/trapperkeeper/internal.clj
```diff
@@ -1,19 +1,23 @@
...
             [puppetlabs.trapperkeeper.common :as common]
             [puppetlabs.trapperkeeper.services :as s]
             [puppetlabs.kitchensink.core :as ks]
             [schema.core :as schema]
             [clojure.core.async :as async]
             [clojure.core.async.impl.protocols :as async-prot]
-            [clojure.string :as string]))
+            [clojure.string :as string]
+            [me.raynes.fs :as fs]
+            [clojure.string :as str]))
```
haus: `clojure.string` is already required as `string`, so requiring it again doesn't seem necessary.

janelu2: wasn't me :^) but i'll remove it

src/puppetlabs/trapperkeeper/internal.clj
```diff
@@ -562,6 +586,8 @@
         this)
       (a/start [this]
         (run-lifecycle-fns app-context s/start "start" ordered-services)
+        (let [restart-file (get-in-config (a/get-service this :ConfigService) [:global :restart-file])]
```
haus: If you use `when-let` instead of `let` here, it will only fire when there is a restart-file defined.

janelu2: wow

src/puppetlabs/trapperkeeper/internal.clj
```diff
@@ -44,6 +48,26 @@
    (every? keyword? (keys service-graph))
    (every? (some-fn ifn? service-graph?) (vals service-graph))))
 
+(defn inc-restart-counter!
+  "Increments the counter in a restart-file for purposes of supporting HUP behavior"
+  [restart-file-path]
+  (when (some? restart-file-path)
```
haus: It looks like restart-file is not a collection, but is going to be a string? I think if you make the change I suggested in start to use `when-let` instead of let, you can remove this `when` call.

src/puppetlabs/trapperkeeper/internal.clj
```diff
@@ -44,6 +47,26 @@
...
+    (try
+      (if (fs/exists? restart-file-path)
+        (spit restart-file-path (inc (Long/parseLong (slurp restart-file-path))))
+        (let [path-list (string/split restart-file-path #"/")
```
haus: if you want the dirname of the restart-file, try `fs/parent restart-file-path`.

```
(fs/parent "/Users/matthaus/foo/bar/baz")
=> #object[java.io.File 0x2fa0d4fd "/Users/matthaus/foo/bar"]
```

src/puppetlabs/trapperkeeper/internal.clj
```diff
@@ -44,6 +47,26 @@
    (every? keyword? (keys service-graph))
    (every? (some-fn ifn? service-graph?) (vals service-graph))))
 
+(defn inc-restart-counter!
+  "Increments the counter in a restart-file for purposes of supporting HUP behavior"
+  [restart-file-path]
+  (when (some? restart-file-path)
+    (try
+      (if (fs/exists? restart-file-path)
```
haus: May be worth adding an `fs/readable?` call somewhere in here.

test/puppetlabs/trapperkeeper/services_test.clj
```diff
@@ -62,6 +65,50 @@
...
+(deftest test-restart-file-correct-count
+  (testing "create a restart file and check that it correctly increments when a service starts"
+    (let [service1 (service Service1
+                            []
+                            (service1-fn [this] "hi"))
+          service2 (service Service1
+                            []
+                            (service1-fn [this] "hi"))]
+          (with-app-with-config app [service1] {:global {:restart-file "./dev-resources/counter"}}
+            (is (some? app))
```
haus: the test above this uses `(is (not (nil? app)))` instead, which is more intuitive.

src/puppetlabs/trapperkeeper/internal.clj
```diff
@@ -44,6 +47,26 @@
...
+              dir-path (string/join "/" (drop-last path-list))]
+          (when (not-empty (drop-last path-list))
+            (fs/mkdirs dir-path))
+          (spit restart-file-path "0")
+          (inc-restart-counter! restart-file-path)))
+      (catch ArithmeticException e
+        (spit restart-file-path "0")
+        (log/warn "Number of restarts has exceeded Long/MAX_VALUE, resetting file to 0"))
+      (catch NumberFormatException e
+        (log/error "Invalid counter; restart file must only consist of one number with no whitespace")))))
```
haus: Why are you only resetting the file to 0 in one of the error cases?

test/puppetlabs/trapperkeeper/services_test.clj
janelu2: I was thinking of actually taking that line completely out since it is tested in the above test and is not relevant to this test

haus: That seems reasonable to me. If it has gotten into the body of the macro, there should be an app.

src/puppetlabs/trapperkeeper/internal.clj
janelu2: that is when overflow occurs

haus:
@janelu2 I left a few suggestions for different clojure functions to use. This looks pretty reasonable to me overall.

src/puppetlabs/trapperkeeper/internal.clj
haus: Right, i understand that you're resetting it when it overflows. I'm wondering why not also set it to 0 when the file is unparseable for some reason?

janelu2: OH that's a good point. i'll add that to the second exception catch

janelu2: maybe it could just use a `finally`?

haus: `finally` will *always* run, regardless of whether an exception was thrown, so I don't think you want to use that here. https://clojuredocs.org/clojure.core/finally

janelu2: oooo gotcha. thanks!

src/puppetlabs/trapperkeeper/internal.clj
```diff
@@ -44,6 +47,25 @@
...
+      (if (and (fs/writeable? restart-file-path) (fs/readable? restart-file-path))
+        (spit restart-file-path (inc (Long/parseLong (slurp restart-file-path))))
+        (log/error "Restart File is not readable and/or writeable"))
+      (let [dir-path (fs/parent restart-file-path)]
+        (fs/mkdirs dir-path)
+        (spit restart-file-path "0")
+        (inc-restart-counter! restart-file-path)))
+    (catch ArithmeticException e
+      (spit restart-file-path "0")
+      (log/warn "Number of restarts has exceeded Long/MAX_VALUE, resetting file to 0"))
```
haus: Should the file be reset to 0, or set to 1?

janelu2: hmmmmmmmmmmmmmmmmmm i was thinking about this as well and could not decide. would it be bad if we skipped 0? jumping to 1 might be intuitive since it means a service has started once. i feel iffy about skipping 0.. what is your reasoning on this?

src/puppetlabs/trapperkeeper/internal.clj
```diff
@@ -44,6 +47,25 @@
...
+  "Increments the counter in a restart-file for purposes of supporting HUP behavior"
+  [restart-file-path]
+  (try
+    (if (fs/exists? restart-file-path)
+      (if (and (fs/writeable? restart-file-path) (fs/readable? restart-file-path))
+        (spit restart-file-path (inc (Long/parseLong (slurp restart-file-path))))
+        (log/error "Restart File is not readable and/or writeable"))
+      (let [dir-path (fs/parent restart-file-path)]
+        (fs/mkdirs dir-path)
+        (spit restart-file-path "0")
```
haus: Any reason not to just spit "1" here?

src/puppetlabs/trapperkeeper/internal.clj
haus: not really, just in the initialization case you're setting it to 1, so not sure what the difference here is.

janelu2: ok that's what i thought, wasn't sure about it

test/puppetlabs/trapperkeeper/services_test.clj
```diff
@@ -5,13 +5,16 @@
...
+              with-app-with-config]]
             [schema.test :as schema-test]
             [puppetlabs.kitchensink.testutils.fixtures :refer [with-no-jvm-shutdown-hooks]]
             [puppetlabs.trapperkeeper.internal :as internal]
             [puppetlabs.trapperkeeper.testutils.logging :refer [with-test-logging]]
             [puppetlabs.trapperkeeper.core :as tk]
-            [puppetlabs.kitchensink.testutils :as ks-testutils])
+            [puppetlabs.kitchensink.testutils :as ks-testutils]
+            [me.raynes.fs :as fs]
+            [clojure.string :as str])
```
haus: This looks unused.

src/puppetlabs/trapperkeeper/internal.clj
janelu2: because it gets thrown back into inc-restart-counter!

test/puppetlabs/trapperkeeper/services_test.clj
janelu2: the `with-test-logging` is unused too, should I remove that as well?

haus: sure.

test/puppetlabs/trapperkeeper/services_test.clj
```diff
@@ -62,6 +65,55 @@
           app (bootstrap-services-with-empty-config [service1])]
       (is (not (nil? app))))))
 
+(deftest test-restart-file-correct-count
+  (testing "create a restart file and check that it correctly increments when a service starts"
+    (let [service1 (service Service1
```
haus: kitchensink has a temp-file function that may be helpful here. in the let block, you can assign the results of a `ks/temp-file` call to a `restart-file` var, and then use it below in your service configs. Then you don't need to worry about deleting it when the test is over.

That should also remove the need to require the fs library for your tests.

src/puppetlabs/trapperkeeper/internal.clj
haus: but if you spit "1", you don't need to call `inc-restart-counter!` again, you can just exit the function.

janelu2: i guess now that i've changed up some things, it doesn't necessary need to be thrown back in

janelu2: changing that now

src/puppetlabs/trapperkeeper/internal.clj
```diff
@@ -44,6 +47,25 @@
...
+        (inc-restart-counter! restart-file-path)))
+    (catch ArithmeticException e
+      (spit restart-file-path "1")
+      (log/warn "Number of restarts has exceeded Long/MAX_VALUE, resetting file to 0"))
```
haus: These log messages should be updated now that it's being reset to 1 and not 0.

src/puppetlabs/trapperkeeper/internal.clj
```diff
@@ -44,6 +47,25 @@
...
+    (catch NumberFormatException e
+      (spit restart-file-path "1")
+      (log/error "Restart file is unparesable, resetting file to 0"))))
```
haus: unparesable => unparseable

src/puppetlabs/trapperkeeper/internal.clj
janelu2: haha yeah

src/puppetlabs/trapperkeeper/internal.clj
janelu2: :scream:

src/puppetlabs/trapperkeeper/internal.clj
```diff
@@ -44,6 +47,25 @@
...
    (every? (some-fn ifn? service-graph?) (vals service-graph))))
 
+(defn inc-restart-counter!
+  "Increments the counter in a restart-file for purposes of supporting HUP behavior"
+  [restart-file-path]
+  (try
+    (if (fs/exists? restart-file-path)
+      (if (and (fs/writeable? restart-file-path) (fs/readable? restart-file-path))
+        (spit restart-file-path (inc (Long/parseLong (slurp restart-file-path))))
+        (log/error "Restart File is not readable and/or writeable"))
```
haus: May be useful to include the file we can't read/write here, as we know what it is.

haus:
@janelu2 pass number two complete. looking super good.

src/puppetlabs/trapperkeeper/internal.clj
```diff
@@ -44,6 +47,24 @@
...
    (every? (some-fn ifn? service-graph?) (vals service-graph))))
 
+(defn inc-restart-counter!
+  "Increments the counter in a restart-file for purposes of supporting HUP behavior"
+  [restart-file-path]
+  (try
+    (if (fs/exists? restart-file-path)
+      (if (and (fs/writeable? restart-file-path) (fs/readable? restart-file-path))
+        (spit restart-file-path (inc (Long/parseLong (slurp restart-file-path))))
+        (log/error (format "Restart file, %s, is not readable and/or writeable" restart-file-path)))
```
cprice404: Wondering if we want this to be a fatal error?  It kind of seems like it might need to be.

src/puppetlabs/trapperkeeper/internal.clj
```diff
@@ -44,6 +47,24 @@
...
+      (let [dir-path (fs/parent restart-file-path)]
+        (fs/mkdirs dir-path)
+        (spit restart-file-path "1")))
+    (catch ArithmeticException e
+      (spit restart-file-path "1")
+      (log/warn "Number of restarts has exceeded Long/MAX_VALUE, resetting file to 1"))
```
cprice404: I'd probably log this at DEBUG level.

src/puppetlabs/trapperkeeper/internal.clj
janelu2: what do we call to make this a fatal error?

cprice404: something like:

```clj
(throw (IllegalStateException. (format "Restart file ... ")))
```

The exception will get logged, so you can probably just replace your 'log' line with that.

test/puppetlabs/trapperkeeper/services_test.clj
cprice404: +1 on using temp files for tests.

src/puppetlabs/trapperkeeper/internal.clj
```diff
@@ -562,6 +583,8 @@
         this)
       (a/start [this]
         (run-lifecycle-fns app-context s/start "start" ordered-services)
+        (when-let [restart-file (get-in-config (a/get-service this :ConfigService) [:global :restart-file])]
+          (inc-restart-counter! restart-file))
```
haus: I can't comment effectively on the line in question, but I think you'll *also* want to call this function below in the `a/restart` function.

test/puppetlabs/trapperkeeper/services_test.clj
```diff
@@ -62,6 +63,55 @@
...
+(deftest test-invalid-restart-file
+  (testing "restart file should reset to 0 if the file is unparesable"
+    (let [service1 (service Service1
+                            []
+                            (service1-fn [this] "hi"))]
+      (spit "./invalid" "hello")
+      (with-app-with-config app [service1] {:global {:restart-file "./invalid"}}
+        (= (slurp "./invalid") "0"))
+       (fs/delete "./invalid"))))
+
```
cprice404: Can we add a few tests for things like:

1. Some portion of parent directory structure doesn't exist
2. File is not writable and/or readable

test/puppetlabs/trapperkeeper/services_test.clj
```diff
@@ -62,6 +63,55 @@
...
 
+(deftest test-restart-file-correct-count
+  (testing "create a restart file and check that it correctly increments when a service starts"
+    (let [service1 (service Service1
+                            []
+                            (service1-fn [this] "hi"))
+          service2 (service Service1
+                            []
+                            (service1-fn [this] "hi"))]
+          (with-app-with-config app [service1] {:global {:restart-file "./dev-resources/counter"}}
```
haus: If you add the function to the `a/restart` lifecycle function, I would remove the second service from this test and just call restart on the app, and then assert that the count had been incremented.

src/puppetlabs/trapperkeeper/internal.clj
cprice404: That sounds right (and, good catch!)... but, I'm a little confused because it appears that there is a test that is asserting that the counter gets incremented due to a restart, and that test is passing?

test/puppetlabs/trapperkeeper/services_test.clj
cprice404: Isn't that what the next test does?

src/puppetlabs/trapperkeeper/internal.clj
janelu2: correct me if i'm wrong, but i thought that it was sufficient because `restart` calls `start`

haus: I don't think the current tests have any assertions. They just have `(= thing thing)` not `(is (= thing thing))`.

test/puppetlabs/trapperkeeper/services_test.clj
cprice404: It does seem like the second service is not needed for this test, though.

test/puppetlabs/trapperkeeper/services_test.clj
```diff
@@ -62,6 +63,55 @@
...
+            (= (slurp "./dev-resources/counter") "1"))
```
haus: For tests, you'll want to use the `is` function around any of your equality tests to ensure that you are asserting equality and not just calling the `=` function.

src/puppetlabs/trapperkeeper/internal.clj
cprice404: @haus haha, um, yeah, that would explain it.

test/puppetlabs/trapperkeeper/services_test.clj
cprice404: As @haus mentioned, these will all need to be changed to `(is (= (...)))`.

test/puppetlabs/trapperkeeper/services_test.clj
```diff
@@ -62,6 +63,55 @@
...
+      (fs/delete "./dev-resources/counter"))))
+
+(deftest test-restart-file-big-int-limit
+  (testing "restart file should reset to 0 if the counter is too large"
+    (let [service1 (service Service1
+                            []
+                            (service1-fn [this] "hi"))
+          service2 (service Service1
+                            []
+                            (service1-fn [this] "hi"))]
```
cprice404: Probably don't need the second service here either.

test/puppetlabs/trapperkeeper/services_test.clj
haus: Oh yea, i definitely missed the test that explicitly called restart. cool.

janelu2:
ok @cprice404 added 2 more tests and did some cleanup, take a look?

test/puppetlabs/trapperkeeper/services_test.clj
```diff
@@ -62,6 +64,67 @@
...
+  (testing "create a restart file and check that it correctly increments when a service starts"
+    (let [service1 (service Service1
+                            []
+                            (service1-fn [this] "hi"))
+          temp-file (kitchensink/temp-file-name "counter")]
+          (with-app-with-config app [service1] {:global {:restart-file temp-file}}
+            (app/restart app)
+            (is (= (slurp temp-file) "2"))))))
+
+(deftest test-restart-file-HUP-restart
```
haus: This test looks identical to the one above it?

cprice404:
minor suggestion for future PRs:

Once folks have started reviewing your PR, if you can do your subsequent changes as just additional, new commits instead of squashing / rebasing, it makes it easier to do a targeted review of just the things that were commented on in the original PR review.  In some cases maybe we'll squash/rebase *after* the review is finished, but in some cases it may also be fine to just leave the extra commits in the history.

Not a huge deal, just something to consider for future PRs.

I'll review again now.

test/puppetlabs/trapperkeeper/services_test.clj
```diff
@@ -62,6 +64,67 @@
...
+                            (service1-fn [this] "hi"))
+          temp-file (kitchensink/temp-file-name "counter")]
+      (spit temp-file "9223372036854775807")
+      (with-app-with-config app [service1] {:global {:restart-file temp-file}}
+        (is (= (slurp temp-file) "1")))
+      (with-app-with-config app [service1] {:global {:restart-file temp-file}}
+        (is (= (slurp temp-file) "2"))))))
+
+(deftest test-invalid-restart-file
+  (testing "restart file should reset to 1 if the file is unparesable"
```
haus: unparesable => unparseable

test/puppetlabs/trapperkeeper/services_test.clj
janelu2: shoot, that (app/restart app) should not be there. sorry a bit out of a it this morning

test/puppetlabs/trapperkeeper/services_test.clj
```diff
@@ -62,6 +64,67 @@
...
+      (fs/chmod "u-rw" (fs/touch temp-file))
+      (is (thrown? IllegalStateException
+                   (with-app-with-config app [service1] {:global {:restart-file temp-file}}))))))
+
+(deftest test-restart-file-missing-parent-dirs
+  (testing "some portion of the parent directory structure does not exist"
+    (let [service1 (service Service1
+                            []
+                            (service1-fn [this] "hi"))
+          temp-file (fs/file (kitchensink/temp-dir "foo") "bar" "baz" "counter")]
```
haus: Would it be worth adding an assertion here that temp-file doesn't exist?
`(is (not (fs/exists? temp-file)))`

src/puppetlabs/trapperkeeper/internal.clj
```diff
@@ -572,6 +595,8 @@
           (doseq [svc-id (keys services-by-id)] (swap! app-context assoc-in [:service-contexts svc-id] {}))
           (run-lifecycle-fns app-context s/init "init" ordered-services)
           (run-lifecycle-fns app-context s/start "start" ordered-services)
+          (when-let [restart-file (get-in-config (a/get-service this :ConfigService) [:global :restart-file])]
+            (inc-restart-counter! restart-file))
```
cprice404: I'm not 100% sure what my opinion is on this comment I'm about to make :)  but, just for consideration:

It might be a little more DRY if we moved the `when-let` part into the body of your function, and just passed in `this` as an extra argument to that function?

test/puppetlabs/trapperkeeper/services_test.clj
janelu2: i had that in my test, wasn't sure if i should have kept it. i'll put it back in

test/puppetlabs/trapperkeeper/services_test.clj
```diff
@@ -62,6 +64,67 @@
           app (bootstrap-services-with-empty-config [service1])]
       (is (not (nil? app))))))
 
+(deftest test-restart-file-correct-count
+  (testing "create a restart file and check that it correctly increments when a service starts"
+    (let [service1 (service Service1
+                            []
+                            (service1-fn [this] "hi"))
+          temp-file (kitchensink/temp-file-name "counter")]
```
cprice404: Just out of curiosity - did you find a reason to prefer `temp-file-name` here instead of just `temp-file`?  Was it maybe that `temp-file` would actually create the file and you didn't want it to do that?

janelu2: yes, `temp-file-name` does not create a file, only generates a name in tmp/ . Also temp-file did not delete the files off my system after the JVM stopped. very mysterious

cprice404: hrm.  We should probably look into that... but it's not a blocker for this PR, obviously.  And `temp-file-name` seems fine for your use case.

test/puppetlabs/trapperkeeper/services_test.clj
janelu2: oops i did it again

janelu2:
@cprice404 made moar changes

test/puppetlabs/trapperkeeper/services_test.clj
cprice404: I'd also assert that the parent dir doesn't exist.

cprice404:
I suggested one more minor assertion to add, otherwise I think I'm +1 on this now.

janelu2:
@cprice404 MOAR CHANGES

janelu2:
@cprice404 i can't check the temp-dir since it does exist, so i'm just checking one of the other directories that were passed in as an arg to file

cprice404:
@janelu2 yep, that's perfect, thanks.

cprice404:
+1

cprice404:
@haus when you have a computer again, let me know if you have anything else on this one.

PR #249, opened 2016-08-22
(TK-397) Update to latest version of logback

cprice404:
This update bumps us to the latest version of logback, which includes
(among other things) support for adding thread IDs to the HTTP
access log files.


KevinCorcoran:
👍

camlow325:
:shipit:

PR #252, opened 2016-09-22
(TK-405) Support specifying restart-file via a CLI arg

camlow325:
This PR adds the ability for the restart-file setting to be
specified via a CLI arg.  The commit preserves the ability to source the
restart-file setting from the global section in Trapperkeeper
configuration.  In the event that the setting were specified in both
places, the setting used for the CLI arg would be the one used.

This PR also includes some documentation for the restart-file feature.

### DIFF
--- a/documentation/Command-Line-Arguments.md
+++ b/documentation/Command-Line-Arguments.md
@@ -8,11 +8,12 @@ Trapperkeeper's default mode of operation is to handle the processing of applica
 
 Note that if you absolutely need control over the command line argument processing, it is possible to circumvent the built-in handling by calling Trapperkeeper's `bootstrap` function directly; see additional details in the [Bootstrapping](Bootstrapping.md) page.
 
-Trapperkeeper supports three command-line arguments:
+Trapperkeeper supports four command-line arguments:
 
 * `--config/-c`: The path to the configuration file or directory. This option is used to initialize the configuration service. This argument is optional; if not specified, Trapperkeeper will act as if you had given it an empty configuration file.
 * `--bootstrap-config/-b`: This argument is optional; if specified, the value should be a path to a bootstrap configuration file, or a comma separated list of files and directories ([see below](#multiple-bootstrap-files)) that Trapperkeeper will use (instead of looking for `bootstrap.cfg` in the current working directory or on the classpath)
 * `--debug/-d`: This option is not required; it's a flag, so it will evaluate to a boolean.  If `true`, sets the logging level to DEBUG, and also sets the `:debug` key in the configuration map provided by the configuration-service.
+* `--restart-file/-r`: This argument is optional; if specified, the value should be a path to a file in which a start counter is incremented each time Trapperkeeper has started all of the services in an application.  See the [Restart File](Restart-File.md) page for additional details.
 
 ### Multiple bootstrap files
 The `--bootstrap-config` argument can be used to specify multiple bootstrap files. This way, a Trapperkeeper app's bootstrap configuration can be split up into multiple locations. You might want to do this to separate logically related services into their own files for instance.
--- a/documentation/Restart-File.md
+++ b/documentation/Restart-File.md
@@ -1,0 +1,64 @@
+# Experimental Feature: Restart File
+
+When using Trapperkeeper apps inside of packages, it is convenient for a service
+framework to have a clear indication as to when all of the Trapperkeeper
+services in the app have been started -- as opposed to just knowing when the
+Java process hosting the app has been spawned.  The "restart file" feature in
+Trapperkeeper provides this capability.  As a reference example, the
+[EZBake](https://github.com/puppetlabs/ezbake) build system for
+Trapperkeeper-based applications makes use of the "restart file" feature.  Its
+service packages can pause a hosting service framework (SysVinit, systemd)
+during a "service start" attempt until the app's services have all been started,
+or have failed to start.
+
+The "restart file" is considered to be a somewhat experimental feature in that
+the implementation may change in a future release.
+
+## Implementation Details
+
+Each time Trapperkeeper has successfully finished processing all of the start
+calls that it makes to each of the services in an application -- both at Java
+process start and after a service reload is requested -- it increments a
+counter in a file on disk.  The location of the file is controlled by the value
+of the `restart-file` setting.
+
+If the value in the file before services are started is '3', for example, the
+value will be updated to '4' after services have been started.  If the file does
+not exist at the time services have been started, the value is written as '1'.
+The value rolls back around to '1' if the value would be incremented beyond the
+maximum value for a `java.lang.Long` or if the contents of the file is otherwise
+unable to be parsed as an integer.
+
+In terms of using the restart file as an indication that services have been
+started -- for example, from a background script that accesses the file in a
+polling loop to determine when the start phase has finished -- it is best to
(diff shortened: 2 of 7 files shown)


src/puppetlabs/trapperkeeper/config.clj
```diff
@@ -59,6 +59,12 @@
     #{".yaml" ".yml"}
     (yaml/parse-string (slurp file))))
 
+(defn override-restart-file-from-cli-data
+  [config-data cli-data]
+  (if-let [cli-restart-file (:restart-file cli-data)]
+    (assoc-in config-data [:global :restart-file] cli-restart-file)
```
cprice404: Do you think it would be worth adding a warning log message here if it's defined in both places?

I don't know if the logging subsystem is even initialized by this point, so, may not be a realistic option.

camlow325: Yeah, a warning might be good here.  I guess even if logback isn't initialized by this point - which I don't know yet - we could still capture the warning somewhere - like to the process' stdout.  So maybe good regardless?

cprice404: (shrug)

camlow325: I added a commit which logs the warning, 2c0648d.  The one somewhat annoying thing about it is that this function ends up getting logged twice at startup because this function is called from two places at startup in the TK code:

https://github.com/puppetlabs/trapperkeeper/blob/1.5.0/src/puppetlabs/trapperkeeper/core.clj#L125

This one is logged before global logback configuration is done by Trapperkeeper.

https://github.com/puppetlabs/trapperkeeper/blob/1.5.0/src/puppetlabs/trapperkeeper/config.clj#L103

This one is logged post Trapperkeeper logback reconfiguration.

Seems like this would come up so seldom that this isn't a big deal but just wanted to point it out.

cprice404:
/nag @jpinsonault @rlinehan , this could use a second review if one of you gets a second.

src/puppetlabs/trapperkeeper/config.clj
rlinehan: So it would get logged in two different places at startup (stdout and the application's configured log with logback)? That seems fine to me, especially since we don't expect it to happen very often.

documentation/Restart-File.md
```diff
@@ -0,0 +1,64 @@
+# Experimental Feature: Restart File
+
+When using Trapperkeeper apps inside of packages, it is convenient for a service
+framework to have a clear indication as to when all of the Trapperkeeper
+services in the app have been started -- as opposed to just knowing when the
+Java process hosting the app has been spawned.  The "restart file" feature in
+Trapperkeeper provides this capability.  As a reference example, the
+[EZBake](https://github.com/puppetlabs/ezbake) build system for
+Trapperkeeper-based applications makes use of the "restart file" feature.  Its
```
rlinehan: This is a total nitpick: I feel like just saying "look at EZBake" is a little mean, since EZBake is pretty complex. Would it be possible to link to one file or to a commit that adds support for this rather than just EZBake as a whole?

src/puppetlabs/trapperkeeper/config.clj
```diff
@@ -59,6 +59,19 @@
...
+(defn override-restart-file-from-cli-data
+  [config-data cli-data]
+  (if-let [cli-restart-file (:restart-file cli-data)]
+    (do
+      (if (get-in config-data [:global :restart-file])
+        (log/warnf
+         (str
+          "restart-file setting specified both on command-line and "
+          "in config file, using command-line value: '%s'")
+         cli-restart-file))
```
rlinehan: I made a comment as part of a review and then deleted the comment but there was a notification that it was made, so ... just in case: I thought that there needed to be a `format` instead of `str` but I missed the call to `log/warnf`, so this should be fine. Also it doesn't look like we've got any i18n stuff going on yet in this repo.

Also github--

src/puppetlabs/trapperkeeper/internal.clj
```diff
@@ -168,15 +168,19 @@
...
                      ["-b" "--bootstrap-config BOOTSTRAP-CONFIG-FILE" "Path to bootstrap config file"]
                      ["-c" "--config CONFIG-PATH"
                       (str "Path to a configuration file or directory of configuration files, "
                            "or a comma-separated list of such paths."
                            "See the documentation for a list of supported file types.")]
-                     ["-p" "--plugins PLUGINS-DIRECTORY" "Path to directory plugin .jars"]]
+                     ["-p" "--plugins PLUGINS-DIRECTORY" "Path to directory plugin .jars"]
+                     ["-r" "--restart-file RESTART-FILE"
+                      (str "Path to a file whose contents is incremented each time all of "
+                           "the configured services have been started.")]]
```
rlinehan: Grammar nitpick: I think this should be "Path to a file whose contents *are* incremented," since it's the contents that are incremented.

documentation/Command-Line-Arguments.md
```diff
@@ -8,11 +8,12 @@ Trapperkeeper's default mode of operation is to handle the processing of applica
...
+Trapperkeeper supports four command-line arguments:
 
 * `--config/-c`: The path to the configuration file or directory. This option is used to initialize the configuration service. This argument is optional; if not specified, Trapperkeeper will act as if you had given it an empty configuration file.
 * `--bootstrap-config/-b`: This argument is optional; if specified, the value should be a path to a bootstrap configuration file, or a comma separated list of files and directories ([see below](#multiple-bootstrap-files)) that Trapperkeeper will use (instead of looking for `bootstrap.cfg` in the current working directory or on the classpath)
 * `--debug/-d`: This option is not required; it's a flag, so it will evaluate to a boolean.  If `true`, sets the logging level to DEBUG, and also sets the `:debug` key in the configuration map provided by the configuration-service.
+* `--restart-file/-r`: This argument is optional; if specified, the value should be a path to a file in which a start counter is incremented each time Trapperkeeper has started all of the services in an application.  See the [Restart File](Restart-File.md) page for additional details.
```
jpinsonault: The way this is phrased makes it sound like it's not TK that's doing the incrementing. I'm having a hard time thinking of a better way to phrase it though. Maybe something like "..a path to a file which TK will update when it has started all the services..."

src/puppetlabs/trapperkeeper/config.clj
camlow325: Yeah, it would be logged in both places.  Not great, but not the end of the world, I suppose.

rlinehan: That almost seems better than logging it in the same place twice? Less redundant, at least. Definitely not the end of the world.

documentation/Restart-File.md
camlow325: Maybe a more specific link would be helpful, not sure?  As it stands, we don't plan on making this an externally configurable feature in EzBake - more just some internal plumbing in the build scripts.  I'm not sure if throwing TK users into the middle of the build scripts without context would be too much more helpful.  I'd also be okay with just removing the EzBake comment altogether but @cprice404 thought there might be value just in mentioning it as an example use case.

camlow325:
@jpinsonault @rlinehan Thanks for the feedback.  I put up a separate PR addressing that since this PR had already been merged.  See #255.

PR #256, opened 2016-09-28
(MAINT) Update the CHANGELOG.md for the 1.5.1 release

camlow325:
This commit updates the CHANGELOG.md file for the 1.5.1 release.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,12 @@
+## 1.5.1
+
+This is a minor feature release
+
+* [TK-405](https://tickets.puppetlabs.com/browse/TK-405) - Add support for
+  specifying the restart file option via a command line argument
+
 ## 1.5.0
+
 This is a feature/bugfix/maintenance release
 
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -1,4 +1,12 @@
+## 1.5.1
```
cprice404: I think I like this better as a 'z' than as a 'y', but, worth raising the question.

camlow325: Yeah, I waffled a bit on it too.  I guess arguably it's a "new feature" and so 'y' may make more sense.  But I tend to think of it as just exposing a new way to configure an already existing feature so not sure the 'y' is totally warranted.

In the end, though, I really don't have a strong opinion about this either.

cprice404: Yeah.  It seems like kind of a silly thing to do a Y for so I'm good with this.  Maybe we can get one more person to weigh in, then :ship:

jpinsonault: I'm good with a z. Strictly speaking it's an additive backwards compatible thing, but eh

PR #265, opened 2017-03-27
(TK-439) Handle exceptions having no type in main

camlow325:
Previously, if an unhandled exception that can be coerced into a map but
that does not have a `:type` key in it were propagated up to
Trapperkeeper's main function, the exception output would be mishandled.
This commit rethrows the original exception if a nil value is found for the
`:type` key in the map, allowing the original exception output to be
displayed properly.

### DIFF
--- a/src/puppetlabs/trapperkeeper/core.clj
+++ b/src/puppetlabs/trapperkeeper/core.clj
@@ -173,10 +173,12 @@
           (internal/parse-cli-args!)
           (run))
       (catch map? m
-        (case (without-ns (:type m))
-          :cli-error (quit 1 (:message m) *err*)
-          :cli-help (quit 0 (:message m) *out*)
-          (throw+)))
+        (let [type (:type m)]
+          (if (keyword? type)
+            (case (without-ns (:type m))
+              :cli-error (quit 1 (:message m) *err*)
+              :cli-help (quit 0 (:message m) *out*))))
+        (throw+))
       (finally
         (log/debug (i18n/trs "Finished TK main lifecycle, shutting down Clojure agent threads."))
         (shutdown-agents)))))


rlinehan:
Are there any tests we could add or tweak for this? (I have no idea what the state of testing of stuff like this in tk is - it seems like it would probably be hard to test but figured I should ask).

camlow325:
A couple of things that I think might make it hard to write Clojure unit tests for this is that the code calls `System/exit` and `shutdown-agents`.  I could move some of the code out into a helper function and pass a mock `quit` function into it from a test, if you think that'd be useful.  Think it's worth it?

rlinehan:
Nah, doesn't seem like it. Thanks for confirming my suspicions though.

src/puppetlabs/trapperkeeper/core.clj
```diff
@@ -173,10 +173,12 @@
           (internal/parse-cli-args!)
           (run))
       (catch map? m
-        (case (without-ns (:type m))
-          :cli-error (quit 1 (:message m) *err*)
-          :cli-help (quit 0 (:message m) *out*)
-          (throw+)))
+        (let [type (:type m)]
+          (if (keyword? type)
```
rlinehan: Is it at all possible there might be a `:type` keyword that's not `cli-error` or `cli-help` and we might want to handle that appropriately?  Idk if that would actually happen/what the whole path to get here is, but it seems like maybe having an `:else` for that case might be reasonable?

camlow325: I think there are some others that could be caught here, yeah.  I guess the question for those would be whether it would be preferable for the user to see a simple message as opposed to a full stack trace with the error.  I think the `cli-error` and `cli-help` ones are caught specifically because it is known that the short message they contain is probably all the user would need in order to understand the text whereas with anything else we don't necessarily know what the user might think is interesting to see so we just let the whole stack trace be dumped out.  If we were to put an `:else` here, what do you think it should have in it?

rlinehan: Oh wait, I think I was misunderstanding why the error was previously happening, which was the precondition of `keyword?` in `without-ns`. Alright, this makes more sense now. I'm good with it as-is. Thanks @camlow325!

PR #266, opened 2017-03-30
(maint) Add log message for finishing lifecycle function

haus:
Previously trapperkeeper would log a message when beginning to run a
lifecycle function, but not when the lifecycle function had returned.
This means that when debugging failures it can be challenging to tell
if a service has died mid-lifecycle function (depending on how much
logging that service has implemented). To alleviate that somewhat, this
commit adds a log message at debug level for when the lifecycle function
has finished.


haus:
This gets a bit noisy at debug. I wonder if the closing message should be at trace instead.

PR #268, opened 2017-09-26
(maint) Fix symbol redef warnings under Clojure 1.9

mullr:


MikaelSmith:
Does this need a bump on clj-kitchensink version to be usable? Or does the clojars package avoid that interpretation issue?

mullr:
You'll need the fixed ks lib to avoid the warnings from that, but it's really just warnings. It should be fine to just let these get into clj-parent the next time it needs to be updated for some other reason.

MikaelSmith:
Ok, well I have a PR up to update clj-parent.

CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+## 1.5.3
```
MikaelSmith: If we're looking to do a release, it looks like a number of things have been merged since 1.5.2: https://github.com/puppetlabs/trapperkeeper/compare/1.5.2...master

PR #270, opened 2018-01-22
(maint) Add plugin jars to clojure's base loader

pcarlisle:
In Java 9 to add a jar to the classpath, we have to add a new
classloader in the hierarchy on the Thread (this is where io/resource
will look) and also to the clojure baseLoader (this is where clojure
will look for actual code). In Java 8 since the base hierarchy was
modifiable we could just change one place and it would be seen
everywhere.

I've also made the kitchensink dependency explicit, since it really does depend on this version number.

### DIFF
--- a/src/puppetlabs/trapperkeeper/plugins.clj
+++ b/src/puppetlabs/trapperkeeper/plugins.clj
@@ -98,7 +98,8 @@
         (do
           (verify-no-duplicate-resources plugins)
           (doseq [jar (jars-in-dir plugins)]
-            (log/info (i18n/trs "Adding plugin .jar to classpath." (.getAbsolutePath jar)))
-            (kitchensink/add-classpath jar)))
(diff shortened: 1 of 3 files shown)


src/puppetlabs/trapperkeeper/plugins.clj
```diff
@@ -98,7 +98,8 @@
         (do
           (verify-no-duplicate-resources plugins)
           (doseq [jar (jars-in-dir plugins)]
-            (log/info (i18n/trs "Adding plugin .jar to classpath." (.getAbsolutePath jar)))
-            (kitchensink/add-classpath jar)))
+            (log/warn (i18n/trs "Adding plugin {0} to classpath." (.getAbsolutePath jar)))
```
MikaelSmith: Why do we want this to be a warning?

pcarlisle: we don't...

PR #273, opened 2018-03-08
(TK-466) Log SIGHUP events at INFO level

Sharpie:
This commit boosts the message logged by the SIGHUP handler from DEBUG to INFO.
This change makes it easy to see from logs when a SIGHUP came into the service.


MikaelSmith:
This sounds useful.

PR #274, opened 2018-05-23
(maint) Remove unnecessary refer-clojure

justinstoller:
We previously did not bring in clojure.core's `boolean?` function so that
we could refer kitchensink's. We now bring in all of kitchensink using
the `ks` namespace. This moves us away from partially refering
clojure.core and using the namespace `ks/boolean?` where appropriate.

A stack trace while debugging dependency upgrades brought me to here,
though on further inspection this was not the cause of the issue.
However, this still seems like a more idiomatic usage, so I thought I
would offer it as a maintenance PR.


MikaelSmith:
A little context if this fixes anything would be nice in the commit message.

justinstoller:
Thanks @MikaelSmith ! I was just pushing this up into a PR quickly so I didn't forget about it later, I've amended the commit message.

justinstoller:
How's this look now?

PR #276, opened 2018-06-26
(MAINT) Add comparason functions to log testutils

austb:
Prior to this commit there was no way to compare log levels
without knowing the underlying log levels were logback.

This commit adds functions to compare log levels.

### DIFF
--- a/test/puppetlabs/trapperkeeper/testutils/logging.clj
+++ b/test/puppetlabs/trapperkeeper/testutils/logging.clj
@@ -42,6 +42,33 @@
    :message (.getMessage event)
    :exception (.getThrowableProxy event)})
 
+(defn log-level-greater-than-or-equal?
+  "Tests if the first log level is greater than the second."
+  [left-level right-level]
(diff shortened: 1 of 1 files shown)


test/puppetlabs/trapperkeeper/testutils/logging.clj
```diff
@@ -42,6 +42,33 @@
...
 
+(defn log-level-greater-than-or-equal?
+  "Tests if the first log level is greater than the second."
+  [left-level right-level]
+  (.isGreaterOrEqual (left-level keyword-levels) (right-level keyword-levels)))
+
+(defn log-level-equal?
+  "Tests if the two log levels given are equal."
+  [left-level right-level]
+  (and (log-level-greater-than-or-equal? left-level right-level)
```
justinstoller: is there just a `.isEqual` that we can use here?

justinstoller:
Sounds like folks decided to go a different route with this work

PR #278, opened 2018-10-30
(PDB-2276) Add ability to boot with a parsed config

robdaemon:
Allow an application to modify the config map after the initial parsing to add
default values, validation and upgrade handling. This allows PuppetDB to remove
its dependency on robert.hooke to modify the behavior of parse-config-data.

This supercedes puppetlabs/trapperkeeper#200

### DIFF
--- a/src/puppetlabs/trapperkeeper/core.clj
+++ b/src/puppetlabs/trapperkeeper/core.clj
@@ -1,4 +1,5 @@
 (ns puppetlabs.trapperkeeper.core
+  (:import (clojure.lang IFn))
   (:require [clojure.tools.logging :as log]
             [slingshot.slingshot :refer [try+ throw+]]
             [puppetlabs.kitchensink.core :refer [without-ns]]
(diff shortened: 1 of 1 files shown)


robdaemon:
I'll put up the corresponding PuppetDB PR shortly so it's more clear how this is used.

src/puppetlabs/trapperkeeper/core.clj
```diff
@@ -143,19 +179,34 @@
...
-  (let [app (boot-with-cli-data cli-data)]
-    ;; This adds the TrapperkeeperApp instance to the tk-apps list, so that
-    ;; it can be referenced in a remote nREPL session, etc.
-    (swap! internal/tk-apps conj app)
-    (internal/register-sighup-handler)
-    (run-app app)
-    (swap! internal/tk-apps (partial remove #{app}))))
+  (bootstrap-with-app (boot-with-cli-data cli-data)))
+
+(schema/defn run-with-config
```
rbrw: Contemplating whether it might be plausible to just add another arity to the existing `run` function instead, i.e.:

```
(schema/defn run
  ;; Update docstring to note that cli-data and config should "agree".
  ([cli-data :- common/CLIData]
   (run cli-data (config/parse-config-data cli-data)))
  ([cli-data :- common/CLIData
    config :- (schema/pred map?)]
   (let [app (boot-with-config cli-data config)]
     ...)))
 ```

File: src/puppetlabs/trapperkeeper/core.clj [hunk 2]
```diff
@@ -93,6 +94,18 @@
    :post [(satisfies? app/TrapperkeeperApp %)]}
   (boot-services-with-config-fn services (constantly config-data)))
 
+(schema/defn boot-with-config-fn :- (schema/protocol app/TrapperkeeperApp)
```
rbrw: Wondering if we might be able to convince ourselves that the config doesn't need to be a function?  (We might not, if it was intended to allow for a dynamic config that could change on say SIGHUP, but I suspect that's  not how you'd ever want to coordinate such a thing.)  And if we can, then perhaps this could just be boot-with-config.

PR #279, opened 2019-01-09
(maint) update clj-parent and nrepl versions

robdaemon:
CIDER now requires a newer nrepl, and org.clojure/tools.nrepl is deprecated

This updates trapperkeeper to newer clj-parent and the new nrepl

Requires: puppetlabs/clj-parent#155

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -8,7 +8,7 @@
 
   :min-lein-version "2.7.1"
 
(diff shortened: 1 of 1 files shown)


robdaemon:
This change is massive. We have tools.nrepl specified all over the place. I'm going to close this for now, and figure out how deep this rabbit hole goes and put a story in JIRA for it.

PR #280, opened 2019-04-17
(orch-2282) Updates trapperkeeper to use the latest nrepl/nrepl

speedofdark:
This is part of a series of changes to support nrepl/nrepl in trapperkeeper. It
needs to be merged after the corresponding changes in clj-parent.

This change requires lein 2.9.0 or later to work; this is when the lein project
changed from the legacy org.clojure/tools.nrepl to the supported nrepl/nrepl
line of development.

This release drops support for JDK7, as nrepl/nrepl has a minimum requirement
of JDK8.

Must not be merged until after https://github.com/puppetlabs/clj-parent/pull/167 is merged.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,14 @@
+## 2.0.0
+
+This is a maintenance release
+
+* [ORCH-2282](https://tickets.puppetlabs.com/browse/ORCH-2282) - Updates to current clj-parent
+to support using nrepl/nrepl
+* Updates required for using nrepl/nrepl; mainline development for nrepl moved from
+org.clojure/tools.nrepl as of the 0.3.x series (last on this line was 0.2.13)
+* Updating to this version of trapperkeeper requires lein >=2.9.0 (:min-lein-version updated)
+* Drops support for JDK7
+
 ## 1.5.6
 
 This is a maintenance release
--- a/project.clj
+++ b/project.clj
@@ -1,14 +1,14 @@
 (def ks-version "2.5.2")
 
-(defproject puppetlabs/trapperkeeper "1.5.7-SNAPSHOT"
+(defproject puppetlabs/trapperkeeper "2.0.0-SNAPSHOT"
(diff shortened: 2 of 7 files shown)


project.clj
```diff
@@ -8,7 +8,7 @@
 
   :min-lein-version "2.7.1"
```
speedofdark: Do we need to bump this to 2.9.0 as thats where nrepl/nrepl was folded into lein?

speedofdark:
The above fail in Travis is because its trying to run Java8 based nrepl/nrepl in a Java7 environment. I'm not sure if this matters?

project.clj
justinstoller: Sure. If we don't we should probably remove the line about it in the CHANGELOG.

CHANGELOG.md
```diff
@@ -1,3 +1,12 @@
+## 1.5.7
+
+This is a maintenance release
+
+* [ORCH-2282](https://tickets.puppetlabs.com/browse/ORCH-2282) - Updates to current clj-parent
```
justinstoller: Though we don't believe this to cause any regressions, or really provide new features, I'd make it a Y since so many dependencies are changing.

speedofdark: Bumped to 1.6.0-SNAPSHOT.

project.clj
speedofdark: Bumped to 2.9.0.

speedofdark:
I need to clarify some circular version references before we merge this.

CHANGELOG.md
caseywilliams: I know we had some confusion on our own side about how development of nrepl had moved form clojure/tools.nrepl to nrepl/nrepl -- I'm not sure how familiar users would be with that change, but as a newcomer I might appreciate a small note about that transition here. I wouldn't block on that though.

speedofdark: Updated!

PR #281, opened 2019-05-15
(maint) Fix some errors being obscured

pcarlisle:
If a slingshot error was thrown with a `:type` value other than `:cli-help` or
`:cli-error`, the original error would be replaced by an error about a missing
clause in the case statement. This allows those errors to just be rethrown.


pcarlisle:
this error was introduced in the master branch and isn't present in 1.1.x

PR #282, opened 2019-05-16
(TK-480) Enable Cider middleware support in the nrepl service

branan:
Cider has a set of nrepl middlewares which it relies on to provide its full functionality. This adds a new configuration option to the nrepl service (`cider-enabled`) which will inject those middlewares. If that option is disabled, the nrepl service will continue to behave as before.

This also updates the exclusions in duplicate-file detection to handle [custom data readers](https://clojuredocs.org/clojure.core/*data-readers*)

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -44,6 +44,7 @@
                  [puppetlabs/kitchensink ~ks-version]
                  [puppetlabs/i18n]
                  [nrepl/nrepl]
+                 [cider/cider-nrepl "0.21.0"]
                  ]
 
   :deploy-repositories [["releases" {:url "https://clojars.org/repo"
--- a/src/puppetlabs/trapperkeeper/plugins.clj
+++ b/src/puppetlabs/trapperkeeper/plugins.clj
@@ -19,7 +19,10 @@
    (not (.startsWith name "META-INF"))
 
     ;; lein includes project.clj ... no thank you
-   (not (= name "project.clj"))))
+   (not (= name "project.clj"))
+
+    ;; *data-readers* extensions can exist in multiple projects
+   (not (= name "data_readers.clj"))))
 
 (defn- handle-duplicate!
   "Helper for `process-file`; handles a found duplicate.  Throws an exception
--- a/src/puppetlabs/trapperkeeper/services/nrepl/nrepl_service.clj
+++ b/src/puppetlabs/trapperkeeper/services/nrepl/nrepl_service.clj
@@ -2,6 +2,7 @@
   (:require
     [clojure.tools.logging :as log]
     [nrepl.server :as nrepl]
+    [cider.nrepl :as cider]
     [puppetlabs.kitchensink.core :refer [to-bool]]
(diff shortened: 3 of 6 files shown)


project.clj
```diff
@@ -44,6 +44,7 @@
                  [puppetlabs/kitchensink ~ks-version]
                  [puppetlabs/i18n]
                  [nrepl/nrepl]
+                 [cider/cider-nrepl "0.21.0"]
```
branan: Should this version number live in clj-parent?

src/puppetlabs/trapperkeeper/plugins.clj
```diff
@@ -19,7 +19,11 @@
    (not (.startsWith name "META-INF"))
 
     ;; lein includes project.clj ... no thank you
-   (not (= name "project.clj"))))
+   (not (= name "project.clj"))
+
+    ;; *data-readers* extensions can exist in multiple projects
+   (not (= name "data_readers.clj"))
+   ))
```
branan: gah, I thought I'd fixed this autoformatting issue. Will push an update

src/puppetlabs/trapperkeeper/services/nrepl/nrepl_service.clj
```diff
@@ -22,22 +23,23 @@
...
 
 (defn- startup-nrepl
   [get-in-config]
-  (let [{:keys [enabled? port bind handler]} (process-config get-in-config)]
+  (let [{:keys [enabled? port bind middlewares cider-middlewares]} (process-config get-in-config)]
     (if enabled?
-      (do (log/info (i18n/trs "Starting nREPL service on {0} port {1}" bind port))
-          (nrepl/start-server :port port :bind bind :handler handler))
+      (let [handler (apply nrepl/default-handler (concat middlewares cider-middlewares))]
+        (do (log/info (i18n/trs "Starting nREPL service on {0} port {1}" bind port))
```
branan: I guess I could inline this into my let with underscores for these steps? I'm not quite sure what's the more idiomatic way to do this

branan:
Looks like I also need to update https://github.com/puppetlabs/trapperkeeper/blob/master/documentation/Configuring-the-nREPL-Service.md

justinstoller:
Is the cider middleware not able to be used with the [middlewares](https://github.com/puppetlabs/trapperkeeper/blob/master/documentation/Configuring-the-nREPL-Service.md#middlewares) config value?

branan:
I think we could avoid the need for the separate config entry, although there are about two dozen middlewares used by cider, so populating the list would be a bit of a hassle compared to a single entry. We'd still need to ensure the cider-nrepl package is included in the jar, so that we can actually reference those middlewares.

project.clj
speedofdark: most likely. non-trivial dance to get it there, tho.

mwaggett:
This is stale, closing.

PR #285, opened 2019-08-09
(maint) clj-parent 4.0 support

justinstoller:


test/puppetlabs/trapperkeeper/core_test.clj
```diff
@@ -45,12 +45,13 @@
                broken-service])))
         (is (logged? #"Error during app buildup!" :error)
             "App buildup error message not logged")))
-    (is (thrown-with-msg?
-         RuntimeException #"Service does not define function 'foo'"
-         (macroexpand '(puppetlabs.trapperkeeper.services/service
-                         puppetlabs.trapperkeeper.core-test/FooService
-                         []
-                         (init [this context] context)))))))
+    (try (macroexpand '(puppetlabs.trapperkeeper.services/service
```
justinstoller: @speedofdark didn't you run into this when you were upgrading tk? I thought there was some discussion about it, but I couldn't remember the specifics....

speedofdark: The errors I'm seeing in Travis don't really look familiar. IS this the wrapping an exception vs. not thing we ran into?

justinstoller: I thought it was related to that (though I might be wrong) do you remember what the outcome of that conversation was?

justinstoller: Sorry, the errors in travis were because the .travis.yml configuration was out of date. I've pushed an update to fix that.

speedofdark: (we ended up unwrapping the exception so it would be able to deal with the raw one vs. the wrapped one, if I recall correctly.)

test/puppetlabs/trapperkeeper/bootstrap_test.clj
```diff
@@ -46,21 +46,24 @@
               (is (= (hello-world hello-world-svc) "hello world")))))
 
         (testing "Gives precedence to bootstrap config in cwd"
```
rbrw: If the new approach is doing what I think it may be, then I wondered if it might be possible, and/or easy enough to test in a temp dir or at least in target/ so we don't have a test clobbering a file in the working tree.  In pdb we have some tests that test the uberjar itself from the outside (we run those separately in travis), which would make that eas(y|ier), but I'm guessing tk doesn't yet.  Might be too difficult for now, but would be nice to adjust if we can.

justinstoller: Hey @rbrw , that's a great idea and when I saw this comment I started work on it. I've run into an issue that's currently blocking me (probably just needs a bit more debugging my system) but I'd like to set it down for now and revisit, so I can get this tk version into our 4.x clj-parent update and get info about it out to dev today. I've filed [TK-483](https://tickets.puppetlabs.com/browse/TK-483) which I'm hoping to do in a couple weeks when I investigate [SERVER-2421](https://tickets.puppetlabs.com/browse/SERVER-2421).

rbrw:
Justin Stoller <notifications@github.com> writes:

> Hey @rbrw , that's a great idea and when I saw this comment I started
> work on it. I've run into an issue that's currently blocking me
> (probably just needs a bit more debugging my system) but I'd like to
> set it down for now and revisit, so I can get this tk version into our
> 4.x clj-parent update and get info about it out to dev today. I've
> filed [TK-483](https://tickets.puppetlabs.com/browse/TK-483) which I'm
> hoping to do in a couple weeks when I investigate
> [SERVER-2421](https://tickets.puppetlabs.com/browse/SERVER-2421).

Sounds good.  Thanks.

PR #287, opened 2020-02-25
(PDB-4636) request-shutdown: support custom exit status/messages

rbrw:
This makes it possible to specify a process exit status value and final messages when calling `request-shutdown`.  The series also fixes some bugs, and adds fully "external" top-level tests.  I needed to make any number of somewhat arbitrary choices, particularly regarding the tests, and mostly leaned toward things previously done in pdb, but I'm happy to rework whatever we like.

You can try all the new tests (which for now just print *everything*, to make CI postmortems easier) by invoking this:
```
lein uberjar
ext/test/run-all
```
from the project directory.

You can also run only the custom exit status tests by invoking `ext/test/custom-exit-status` directly.  And note that particular test, unlike `top-level-cli` doesn't actually rely on the uberjar right now).

Please see the individual commit messages for additional information, and looking at the individual commits might make it easier to become confident in the changes.  I attempted to present the changes step-by-step.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,29 @@
 language: clojure
 lein: 2.9.1
-jdk:
-  - openjdk11
-  - openjdk8
+
+jobs:
+  include:
+    - jdk: openjdk8
+      name: lein test (openjdk8)
+    - jdk: openjdk11
+      name: lein test (openjdk11)
+    - jdk: openjdk8
+      name: external tests (openjdk8)
+      script: lein uberjar && ext/test/run-all
+    - jdk: openjdk11
+      name: external tests (openjdk11)
+      script: lein uberjar && ext/test/run-all
+    - name: external tests (openjdk11)
+      # Apparently travis' lein support is broken right now
+      language: java
+      os: osx
+      osx_image: xcode10.3
+      script: |
+        # prep in standalone script so things like set -x don't affect travis
+        ext/travisci/prep-macos \
+        && export PATH="$(pwd)/ext/travisci/bin:$PATH" \
+        && lein uberjar \
+        && ext/test/run-all
+
 notifications:
   email: false
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,3 +27,8 @@ jobs:
 
 notifications:
   email: false
+
+cache:
+  directories:
+    - $HOME/.m2
+    - $HOME/Library/Caches/Homebrew
--- a/src/puppetlabs/trapperkeeper/core.clj
+++ b/src/puppetlabs/trapperkeeper/core.clj
@@ -173,11 +173,11 @@
           (internal/parse-cli-args!)
           (run))
       (catch map? m
-        (let [type (:type m)]
+        (let [type (:kind m)]
           (if (keyword? type)
-            (case (without-ns (:type m))
-              :cli-error (quit 1 (:message m) *err*)
-              :cli-help (quit 0 (:message m) *out*)
+            (case (without-ns (:kind m))
+              :cli-error (quit 1 (:msg m) *err*)
+              :cli-help (quit 0 (:msg m) *out*)
               ;; By default let the throw below reraise the error
               nil)))
         (throw+))
--- a/src/puppetlabs/trapperkeeper/core.clj
+++ b/src/puppetlabs/trapperkeeper/core.clj
@@ -142,7 +142,9 @@
       (internal/call-error-handler! shutdown-reason)
       (internal/shutdown! (app/app-context app))
       (when-let [error (:error shutdown-reason)]
-        (throw error)))))
+        (throw error))
+      (when (= :requested (:cause shutdown-reason))
+        (select-keys shutdown-reason [::exit])))))
 
 (schema/defn run
   "Bootstraps a trapperkeeper application and runs it.
@@ -155,26 +157,20 @@
     ;; it can be referenced in a remote nREPL session, etc.
     (swap! internal/tk-apps conj app)
     (internal/register-sighup-handler)
-    (run-app app)
-    (swap! internal/tk-apps (partial remove #{app}))))
-
-(def exit-request-schema
-  "A process exit request like
-  {:kind :puppetlabs.trapperkeeper.core/exit
-   :status 7
-   :messages [[\"something for stderr\n\" *err*]]
-              [\"something for stdout\n\" *out*]]
-              [\"something else for stderr\n\" *err*]]"
-  {:kind ::exit
-   :status schema/Int
-   :messages [[(schema/one schema/Str "message")
-               (schema/one java.io.Writer "stream")]]})
+    (let [{:keys [::exit] :as result} (run-app app)]
+      ;; Q: If it's appropriate (or even just acceptable) for this
+      ;; removal to be handled via catch/finally, then it'd be a bit
+      ;; simpler to just throw the exit from run-app.
+      (swap! internal/tk-apps (partial remove #{app}))
+      (when exit
+        (throw (ex-info (i18n/trs "Process exit requested")
+                        (assoc (::exit result) :kind ::exit)))))))
 
 (defn- parse-args [args]
   "Returns valid CLIData or throws an ::exit."
   (letfn [(quit [status msg stream ex-msg]
             (throw (ex-info ex-msg
-                            ;; Matches exit-request-schema
+                            ;; :status and :messages match exit-request-schema
                             {:kind ::exit
                              :status status
                              :messages [[(str msg "\n") stream]]})))]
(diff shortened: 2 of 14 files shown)


src/puppetlabs/trapperkeeper/core.clj
```diff
@@ -154,33 +157,74 @@
     ;; it can be referenced in a remote nREPL session, etc.
     (swap! internal/tk-apps conj app)
     (internal/register-sighup-handler)
-    (run-app app)
-    (swap! internal/tk-apps (partial remove #{app}))))
+    (let [{:keys [::exit] :as result} (run-app app)]
+      ;; Q: If it's appropriate (or even just acceptable) for this
+      ;; removal to be handled via catch/finally, then it'd be a bit
+      ;; simpler to just throw the exit from run-app.
```
Magisus: Not sure about the answer here. Though throwing from `run-app` is consistent with the throw we're already doing for for errors, which could be nice.

Magisus:
This seems like something we'll want to do a Y release for, would you mind updating the SNAPSHOT version in `project.clj` as part of this, to make it easier to release?

Magisus:
I'm trying the new tests now and running into some issues. More details to come.

src/puppetlabs/trapperkeeper/core.clj
mwaggett: I'm into simplicity, so if throwing from `run-app` is simpler..

mwaggett: also curious though: why is the exit being `throw`n anyway? does that not just cause an error? I thought these were intentional exits?

src/puppetlabs/trapperkeeper/core.clj
```diff
@@ -154,33 +157,74 @@
...
+            :cli-help (quit 0 msg *out* (i18n/trs "Command line --help requested"))
+            (throw ex)))))))
+
+(defn- handle-exit
+  "Prints any messages provided to the indicated streams, and returns
+  the appropriate process exit status."
+  [{:keys [status messages]}]
+  ;; Check values carefully here since throwing ::exit is a public interface
+  (letfn [(show [stream msg]
+            (binding [*out* stream] (print msg) (flush)))]
```
mwaggett: nitpick, but I think this might be easier to read/interpret if the functions called within this `binding` were indented underneath, as in the examples here https://clojuredocs.org/clojure.core/binding

src/puppetlabs/trapperkeeper/core.clj
```diff
@@ -154,33 +157,74 @@
...
+    (doseq [[string stream :as m] messages
```
mwaggett: would prefer `message` to `m` here, for readability

src/puppetlabs/trapperkeeper/core.clj
```diff
@@ -154,33 +157,74 @@
...
+            :when (try
```
mwaggett: Why is `:when` a keyword here?

src/puppetlabs/trapperkeeper/internal.clj
```diff
@@ -345,19 +345,35 @@
 ;;;; regarding the cause of the shutdown, and is intended to be passed back
 ;;;; in to the top-level functions that perform various shutdown steps.
 
+(def ^{:doc "A process exit request like
+  {:status 7
+   :messages [[\"something for stderr\n\" *err*]]
+              [\"something for stdout\n\" *out*]]
+              [\"something else for stderr\n\" *err*]]"}
```
mwaggett: From what I've seen, it looks like most functions have doc strings below their declaration, as in `request-shutdown*` below. Why the `{:doc }` here?

mwaggett:
I also had issues running these on OSX (ubuntu-18.04 works fine).
If humans are going to be running these, I feel like it needs to work on OSX, since I think that's our most common dev environment.
But I was curious why this new behavior is being tested in bash scripts, rather than in actual tests, like the rest of this codebase?

src/puppetlabs/trapperkeeper/core.clj
Magisus: `doseq` allows you to put in bindings to filter stuff, see the docs for [for](https://clojuredocs.org/clojure.core/for).

mwaggett: ahh I see, thank you!

src/puppetlabs/trapperkeeper/core.clj
rbrw: @Magisus agreed, and to elaborate, my concern was really just whether there might be some reason TK really meant to avoid the removal on exception -- but I could also imagine it was just an oversight, and that it should always be removed.  If we can convince ourselves that the removal is "fine", then I'd be inclined to switch it to a finally or similar.

rbrw: @mwaggett the `::exit` throws are intentional exits -- see the corresponding catch in `main` below.  We do that so that you can exit "immediately", while still having an opportunity to run any pending handlers, i.e. `finally`, `catch`, etc.

src/puppetlabs/trapperkeeper/core.clj
rbrw: Ahh I sometime use one line for compactness when the body is trivial, but happy to change it. Also prefer the flush on a separate line in this case?

src/puppetlabs/trapperkeeper/core.clj
rbrw: Oh, it's one of the "modifiers" supported by `for`: https://clojuredocs.org/clojure.core/for

src/puppetlabs/trapperkeeper/internal.clj
rbrw: Oh, right, thanks.  I'd just forgotten that `def` also supports a docstring after the name (I believe it didn't in (much?) older clojure versions).  I'll fix it.

src/puppetlabs/trapperkeeper/core.clj
mwaggett: ahh gotcha, that makes sense, thanks!

src/puppetlabs/trapperkeeper/core.clj
mwaggett: yes please. since I'm still getting familiar with Clojure, following standard conventions is helpful to me. 😊

rbrw:
@mwaggett I did it that way because I was trying to test the process exit status, i.e. the jvm exit status, and we have to do that from the outside.  I suppose we could try to do it with clojure.java.shell/sh, but that seems a bit heavyweight just to build/run the ubjerjar, and clojure.java.shell/sh is somewhat limited (i.e. pipes/streaming, managing multiple subprocs, is difficult, etc.).  But we could certainly do it.  Alternately, we could do it from python or ruby if we prefer.  Both still a bit more awkward than the shell for simpler things, but likely better once we get past a certain point of complexity.

Of course, I also proposed this initially just because we already had the code working in pdb for a number of "external" tests.

In any case, if we all decide we'd prefer to do this in some other language, I'd be happy to rework it.  Might need a bit of help if we were to pick ruby, but be good for me to improve there anyway.

rbrw:
@Magisus @mwaggett I'll see if I can figure out the macos failures -- should have added that to the travis matrix anyway, just didn't think of it.  (We've generally supported that for pdb, though recently we've allowed the macos tests to fail after we had some jdk (and/or cert?) related issues, etc. that we didn't want to take the time to resolve -- I forget the details, but may not affect the tests here).

Magisus:
I'm fine with these being bash scripts, at least for now. But I agree they should work on OSX. Do you have a way to test through that, or should someone with a Mac laptop take a look?

rbrw:
I suspect I can fix it (via travis), but if not I'll ping you (and/or Zak may be able to help me) -- thanks.

rbrw:
OK, if the tests pass (macos now included for the external tests), I hope I may have otherwise addressed the current comments.

Magisus:
The hang travis is seeing reproduces for me locally.

rbrw:
Travis actually hung at the same time.  After a restart, it seemed to work fine (offhand, don't think I've changed anything relevant), and Zak and I have tested the current commit in a loop without any trouble.  If I didn't change anything material, then I'm not sure what caused the previous trouble.

Magisus:
My run is no longer hanging either.

.travis.yml
```diff
@@ -1,7 +1,34 @@
...
+      name: lein test (openjdk8)
+    - jdk: openjdk11
+      name: lein test (openjdk11)
+    - jdk: openjdk8
+      name: external tests (openjdk8)
+      script: lein uberjar && ext/test/run-all
+    - jdk: openjdk11
+      name: external tests (openjdk11)
+      script: lein uberjar && ext/test/run-all
+    - name: external tests (openjdk11)
```
mwaggett: nitpick, but should this have a different name than the test above?

.travis.yml
```diff
@@ -1,7 +1,34 @@
...
+      # Apparently travis' lein support is broken right now
```
mwaggett: scripts above are still calling `lein`? I'm a bit confused..

mwaggett:
script works for me now, so I'm good to merge 👍

PR #289, opened 2020-07-24
(TK-487) Allow friendly init/start fail fast via ::exit throw

rbrw:
Allow init and start methods to throw a request-shutdown style ex-info
map to short circuit the startup process and exit with a specified
message and status, rather than a backtrace.

This just provides a short circuiting (immediate) counterpart to the
existing, deferred shutdown requests provided by request-shutdown.

### DIFF
--- a/src/puppetlabs/trapperkeeper/internal.clj
+++ b/src/puppetlabs/trapperkeeper/internal.clj
@@ -176,6 +176,29 @@
         required    []]
     (first (ks/cli! cli-args specs required))))
 
+(def exit-request-schema
+  "A process exit request like
+  {:status 7
+   :messages [[\"something for stderr\n\" *err*]]
+              [\"something for stdout\n\" *out*]]
+              [\"something else for stderr\n\" *err*]]"
+  {:status schema/Int
+   :messages [[(schema/one schema/Str "message")
+               (schema/one java.io.Writer "stream")]]})
+
+(defn exit-exception? [ex]
+  (and (instance? ExceptionInfo ex)
+       (not (schema/check {(schema/optional-key :puppetlabs.trapperkeeper.core/exit)
+                           exit-request-schema}
+                          (ex-data ex)))))
+
+(defn shutdown-reason-for-ex
+  [exception]
+  (if (exit-exception? exception)
+    (merge {:cause :requested}
+           (select-keys (ex-data exception) [:puppetlabs.trapperkeeper.core/exit]))
+    {:cause :service-error :error exception}))
+
 (schema/defn ^:always-validate run-lifecycle-fn!
   "Run a lifecycle function for a service.  Required arguments:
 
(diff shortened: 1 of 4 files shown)


rbrw:
This isn't quite ready (e.g. at least needs tests), but I wanted to see if it seemed plausible before proceeding further.  The current motivation is wanting to be able to shut down in a friendly way when a service detects a problem with the configuration it retrieves via `(get-config)`, without having to carefully coordinate with all other services in order to safely rely on a deferred `request-shtudown`.

src/puppetlabs/trapperkeeper/internal.clj
```diff
@@ -176,6 +176,29 @@
...
+  (and (instance? ExceptionInfo ex)
+       (not (schema/check {(schema/optional-key :puppetlabs.trapperkeeper.core/exit)
+                           exit-request-schema}
+                          (ex-data ex)))))
+
+(defn shutdown-reason-for-ex
+  [exception]
+  (if (exit-exception? exception)
+    (merge {:cause :requested}
+           (select-keys (ex-data exception) [:puppetlabs.trapperkeeper.core/exit]))
```
mwaggett, line 199: Why `select-keys` instead of including all the exception data?
Also, based on your documentation, it looks like `:puppetlabs.trapperkeeper.core/exit` is not a top-level key, it's the value for the `:kind` key. so not sure what the intention was here.

src/puppetlabs/trapperkeeper/internal.clj
```diff
@@ -176,6 +176,29 @@
...
+   :messages [[\"something for stderr\n\" *err*]]
+              [\"something for stdout\n\" *out*]]
+              [\"something else for stderr\n\" *err*]]"
+  {:status schema/Int
+   :messages [[(schema/one schema/Str "message")
+               (schema/one java.io.Writer "stream")]]})
+
+(defn exit-exception? [ex]
+  (and (instance? ExceptionInfo ex)
+       (not (schema/check {(schema/optional-key :puppetlabs.trapperkeeper.core/exit)
```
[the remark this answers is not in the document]
rbrw: Oh, right, because I believe `check` returns `nil` if it matches, info about the mismatch, otherwise.

src/puppetlabs/trapperkeeper/internal.clj
rbrw: If I recall correctly, it's just to be sure that the exception only contributes the key that it's supposed to, e.g. can't clobber other top-level keys, etc.

And I think the idea is that `:kind` is a non-namespaced key that tells you what type of ex-data map you have.  Everything else in the map can be type-specific.  The ':kind` value is namespaced to avoid collisions across namespaces and even across different projects.

We could namespace `:kind`, as say `:puppetlabs/kind`, but I believe there's some precedence at puppet (and I vaguely recall maybe elsewhere) to use `:kind` globally.  TK itself is already using `:kind` for some exit-related code.

PR #297, opened 2021-12-10
Collect exceptions caught during shutdown

### LINKED ISSUE #292: Add option to not swallow exceptions during `shutdown!`
## Use Case

Logging aside, https://github.com/puppetlabs/trapperkeeper/blob/c943510c08acde6b1e3c2b55b450fef325734022/src/puppetlabs/trapperkeeper/internal.clj#L461 swallows any exceptions, which may be OK in production, but in a test suite it can be considered bit of a weakness: it's not desirable to have the test pass with a failing `stop` method.

## Describe the Solution You Would Like

Add an option, e.g. Java system property that allows to re-throw the exception.

One would enable the property in dev/test environments.

DeLaGuardo:
fix #292

### DIFF
--- a/src/puppetlabs/trapperkeeper/internal.clj
+++ b/src/puppetlabs/trapperkeeper/internal.clj
@@ -448,24 +448,34 @@
     (shutdown-on-error [this svc-id f] (shutdown-on-error* shutdown-reason-promise app-context svc-id f))
     (shutdown-on-error [this svc-id f on-error] (shutdown-on-error* shutdown-reason-promise app-context svc-id f on-error))))
 
-(schema/defn ^:always-validate shutdown!
+(schema/defn ^:always-validate shutdown! :- [Throwable]
   "Perform shutdown calling the `stop` lifecycle function on each service,
-   in reverse order (to account for dependency relationships)."
+   in reverse order (to account for dependency relationships).
+   Returns collection of exceptions thrown during shutdown sequence execution."
   [app-context :- (schema/atom a/TrapperkeeperAppContext)]
   (log/info (i18n/trs "Beginning shutdown sequence"))
   (let [{:keys [ordered-services shutdown-channel lifecycle-worker]} @app-context
-        shutdown-fn (fn [] (doseq [[service-id s] (reverse ordered-services)]
-                             (try
-                               (run-lifecycle-fn! app-context s/stop "stop" service-id s)
-                               (catch Exception e
-                                 (log/error e (i18n/trs "Encountered error during shutdown sequence"))))))]
+        result-chan (async/promise-chan (map #(filter (partial not= :ok) %)))
+        shutdown-fn (fn [] (let [results
+                                 (doall
+                                  (map (fn [[service-id s]]
+                                         (try
+                                           (run-lifecycle-fn! app-context s/stop "stop" service-id s)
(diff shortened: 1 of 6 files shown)


src/puppetlabs/trapperkeeper/internal.clj
```diff
@@ -606,8 +616,15 @@
...
-        (shutdown! app-context)
-        this)
+        (a/stop this false))
+      (a/stop [this throw?]
+        (if-let [errors (not-empty (shutdown! app-context))]
+          (let [msg (i18n/trs "Error during app shutdown!")
+                e (ex-info msg {:errors errors})]
+            (log/error e msg)
+            (when throw?
+              (throw e)))
```
frenchy64, line 626: Possible regression: Perhaps `(if thrown? (throw e) this)` to avoid breaking changes.

src/puppetlabs/trapperkeeper/internal.clj
```diff
@@ -448,24 +448,34 @@
...
-                               (run-lifecycle-fn! app-context s/stop "stop" service-id s)
-                               (catch Exception e
-                                 (log/error e (i18n/trs "Encountered error during shutdown sequence"))))))]
+        result-chan (async/promise-chan (map #(filter (partial not= :ok) %)))
+        shutdown-fn (fn [] (let [results
+                                 (doall
+                                  (map (fn [[service-id s]]
+                                         (try
+                                           (run-lifecycle-fn! app-context s/stop "stop" service-id s)
+                                           :ok
```
frenchy64, line 464: Style: Perhaps we could return `nil` here and use `keep` instead of `map`. Maybe then we could remove the `promise-chan` transducer.

src/puppetlabs/trapperkeeper/internal.clj
DeLaGuardo: thanks!

src/puppetlabs/trapperkeeper/internal.clj
DeLaGuardo: I think you are right, it looks much cleaner. Actually I begin with that idea but later my thought drifted towards "results-chan" instead of "errors-chan" as a holder for stuses not only for "failed to stop" services but also for all others, like "service finished but there is a list of pending jobs"

justinstoller (approved):

justinstoller:
Excellent! Thank you both so much for working on this. Would you mind also updating the CHANGELOG file? We can then do a release with these changes on Monday.

DeLaGuardo:
Looking forward for the Monday release :)

Also there was three releases missing in CHANGELOG.md. I added some notes based of comparing the tags but please check them out.

Magisus (approved):

DeLaGuardo:
Hi! @justinstoller, sorry for bothering you. Are you still planning to cut the release any time soon? The release is not urgent; I want to ensure that I didn't miss something obvious in my PR.

Magisus:
Thanks for the reminder, we should get on this. I'll look into it today 🙂

Magisus:
This has been released https://github.com/puppetlabs/trapperkeeper/releases/tag/3.2.0

DeLaGuardo:
Amazing! Thank you very much for the quick response!

PR #309, opened 2023-06-22
(PE-36370) remove yaml support

jonathannewman:
This removes support for yaml configuration files from trapperkeeper.
The yaml configuration is not a common case, and not used in Puppet
Enterprise at all.  The yaml parser is the source of multiple CVEs.


steveax (approved):
👍

PR #310, opened 2023-07-14
(PE-36476) Replace clj-yaml with snakeyaml 2.0

artlawson:
This removes the use of clj-yaml for yaml parsing because internally it uses snakeyaml 1.33. Instead, snakeyaml 2.0 is used directly and a routine for mapping from java structures to clojure structures to clojure structures is added.

### DIFF
--- a/src/puppetlabs/trapperkeeper/common.clj
+++ b/src/puppetlabs/trapperkeeper/common.clj
@@ -1,5 +1,7 @@
 (ns puppetlabs.trapperkeeper.common
-  (:require [schema.core :as schema]))
+  (:require [schema.core :as schema])
+  (:import (java.util List Map Set)
+          (org.yaml.snakeyaml Yaml)))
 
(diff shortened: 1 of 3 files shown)


src/puppetlabs/trapperkeeper/common.clj
```diff
@@ -1,5 +1,8 @@
 (ns puppetlabs.trapperkeeper.common
-  (:require [schema.core :as schema]))
+  (:require [schema.core :as schema])
+  (:import (java.util List Map Set)
+          (java.util.concurrent TimeUnit)
```
jonathannewman, line 4: `TimeUnit` doesn't seem to be used?

jonathannewman (changes requested):
One minor nit, but otherwise looks good to me:

```
Ran 90 tests containing 371 assertions.
0 failures, 0 errors.
```

jonathannewman (approved):

PR #317, opened 2024-06-28
(maint) update logged? for better output, prepare for release

jonathannewman:
Update the `logged?` function to only emit messages when there are
multiple statements when they aren't expected, and also clean up
the output to be more useful.


jonathannewman:
Continuing to ignore the linter issues on this branch.

steveax (approved):

